Issue #3, opened 2018-01-30
Helper function


asafyish:
Hi,

Lets say I have a base32 string 'nbswy3dpo5xxe3de', I was expecting this library to support it as input or at least provide with an helper function to do so, but it seems I need a pre-step to convert the string to a special vec with integers ranging from 0 to 31.

Does it makes sense to add this helper function ?

clarkmoody:
Try using 
```rust
let decode_result = Bech32::from_string("nbswy3dpo5xxe3de".to_string())
```

asafyish:
That's for decoding, which I am able to do.
I am having trouble encoding,
Had to use convert_bits from https://github.com/petertodd/rust-bech32/blob/master/src/wit_prog.rs,
something like:

```rust
    let public_key = "d1jprv3fexqq4v34".to_string().into_bytes();
    let public_key = convert_bits(public_key, 8, 5, true).unwrap();
    println!("public key {:?}", public_key);

    let b = Bech32 {
        hrp: "poc".to_string(), 
        data: public_key
    };
    let encoded = b.to_string().unwrap();
```

maybe convert_bits should be provided in this lib ?

clarkmoody:
I wrapped that functionality in the [bitcoin-bech32](https://crates.io/crates/bitcoin-bech32) create, though `convert_bits` is private there. It does make sense to have `convert_bits` live within this module. I'll work on porting it over from `bitcoin-bech32`.

clarkmoody:
Resolved with #9

PR #4, opened 2018-03-07
Use `FromStr` trait for parsing

sgeisler:
Using `FromStr` potentially saves one `clone()` since the bech32 input string doesn't have to be moved into the `from_string` function. Since the string is never consumed inside the function this should not change the it's behavior.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -39,7 +39,7 @@
 //! let encoded = b.to_string().unwrap();
 //! assert_eq!(encoded, "bech321qpz4nc4pe".to_string());
 //! 
-//! let c = Bech32::from_string(encoded);
+//! let c = encoded.parse::<Bech32>();
 //! assert_eq!(b, c.unwrap());
 //! ```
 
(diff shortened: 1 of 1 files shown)


clarkmoody:
This looks great, thanks! More idiomatic Rust is always appreciated.

I'll need to see how this affects the sister [rust-bech32-bitcoin](https://github.com/clarkmoody/rust-bech32-bitcoin) library before pushing an update to Crates.io.

I'm also working on adding `convert_bits` to this library, according to #3

sgeisler:
> I'll need to see how this affects the sister rust-bech32-bitcoin library before pushing an update to Crates.io.

Maybe we should have a look at other API-breaking changes that might be desirable first. I'll open an issue to collect API cleanup ideas.

> I'm also working on adding convert_bits to this library, according to #3

I'm looking forward to it :) I could use it to get rid of a horribly inefficient implementation of `base32_to_base256` (I'm implementing BOLT 11 payment parsing)

PR #6, opened 2018-03-07
Implement `Display` trait

sgeisler:
- [x] refactor and move `to_string` into `Display` trait
- [x] test `new` function and getters

PR #9, opened 2018-03-09
Add convert bits

clarkmoody:
Imports the `convert_bits` function from `rust-bech32-bitcoin`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -19,26 +19,26 @@
 // THE SOFTWARE.
 
 //! Encoding and decoding Bech32 format
-//! 
-//! Bech32 is a 5-bit (base-32) encoding scheme that produces strings that comprise 
-//! a human-readable part, a separator, a data part, and a checksum. The encoding 
-//! implements a BCH code that guarantees error detection of up to four characters 
+//!
+//! Bech32 is a 5-bit (base-32) encoding scheme that produces strings that comprise
+//! a human-readable part, a separator, a data part, and a checksum. The encoding
+//! implements a BCH code that guarantees error detection of up to four characters
 //! with less than 1 in 1 billion chance of failing to detect more errors.
-//! 
+//!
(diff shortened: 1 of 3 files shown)


clarkmoody:
@sgeisler @apoelstra Any comments on this one?

apoelstra:
Would it make sense to add `BitcoinConversionError` to the top-level `Error` enum?

clarkmoody:
Yep. We can reuse `Error::InvalidData(u8)` and add `InvalidPadding` to the existing enum.

apoelstra:
Great, if you can do that then ACK from me.

apoelstra:
ACK

clarkmoody:
Resolves #3

sgeisler:
I'd like to change the type of `data` to `&[u8]`, so the signature becomes  `fn convert_bits(data: &[u8], from: u32, to: u32, pad: bool)`. I will include this in #6 after rebasing on master.

src/lib.rs
```diff
@@ -287,15 +290,63 @@ impl error::Error for Error {
...
+///
+/// # Examples
+///
+/// ```rust
+/// use bech32::convert_bits;
+/// let base5 = convert_bits(vec![0xff], 8, 5, true);
+/// assert_eq!(base5.unwrap(), vec![0x1f, 0x1c]);
+/// ```
+pub fn convert_bits(data: Vec<u8>, from: u32, to: u32, pad: bool) -> ConvertResult {
+    if from > 8 || to > 8 {
```
sgeisler: What happens if from or to is 0? From looking at the code the function could run indefinitely. Maybe it should panic instead.

apoelstra: Yeah, panicking would be better.

clarkmoody: `panic` on 0 would need a couple additional tests as well

sgeisler: opened #11 to fix this, but it would be nice if #10 could be merged first since rebasing is easier for me this way.

PR #11, opened 2018-03-09
Make convert_bits panic if from or to is 0

sgeisler:
+ add test for `from`/`to` `>8` or `==0` (needs rebase after #10 )

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -328,7 +328,7 @@ type ConvertResult = Result<Vec<u8>, Error>;
 ///
 /// # Panics
 /// Function will panic if attempting to convert `from` or `to` a bit size that
-/// is larger than 8 bits.
+/// is larger than 8 bits or 0 bits.
 ///
 /// # Examples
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -325,7 +325,7 @@ type ConvertResult = Result<Vec<u8>, Error>;
 ///
 /// # Panics
 /// Function will panic if attempting to convert `from` or `to` a bit size that
-/// is larger than 8 bits.
+/// is larger than 8 bits or 0 bits.
```
clarkmoody: Nit: I think it reads better as `is 0 bits or larger than 8 bits.`

sgeisler:
I did the rebase, you can merge it now if you want.

src/lib.rs
sgeisler: Yes, I thought about writing `exactly 0` and it felt a bit over the top, but just changing the order makes it quite clear.

PR #12, opened 2018-03-09
Derive Clone and Copy trait for Error

sgeisler:
Otherwise every other error type encapsulating `Error` will not be able to derive `Clone` and `Copy`.


clarkmoody:
Simple enough 😄

clarkmoody:
(sorry, this did a rebase-and-merge, same as for #7)

apoelstra:
Annoyingly, last I checked `std::io::Error` does not implement `Clone` or `Copy`. I think because this library has no side-effects this is fine, but one reason not to implement these is to future-proof against the future need to add `std::io::Error` to the enum.

To be clear, I think it's super irritating that `std::io::Error` is like this and I'm glad that bech32's Error type is better :). Just a heads up.

PR #14, opened 2018-03-21
Minor nits to get compilation on rustc 1.14 (shipping Debian version)

apoelstra:


apoelstra:
Will give @clarkmoody a couple days to check this out, since it is his library

PR #17, opened 2018-04-06
Add u5 type and traits

### LINKED ISSUE #16: Define u5 type to distinguish between 5bit and 8bit slices
In rust-lightning-invoice I pass around slices of the data part as `&[u8]`, but each byte is guaranteed to be in the range `0..32`. I'd like to encode this guarantee in the type by defining `pub struct u5(u8)`.

The traits `FromBase32` and `IntoBase32` could be introduced and implemented for `&[u8]` and `&[u5]` respectively and replace `convert_bits` (`From<&[u8]> for &[u5]` and `From<&[u5]> for &[u8]` can't be implemented since type and trait are foreign and it would be quite ambiguous). Furthermore  `FromBase32` could be implemented for data structures that get parsed from bech32 data (like lightning invoice fields).

The best case UX would be that you can now implement `FromBase32` for your types and don't have to worry about out-of-range errors anymore. The worst case UX would probably be that every use of an `u5` (formerly `u8`) will now require an `.into::<u8>()`.

What do you think about this proposal @clarkmoody and @apoelstra (I know I brought this up already in #5 but it got lost when we released the new version)?

sgeisler:
Fixes #16. I'm not really sure if the API is as good as it could be. I will have a look at rust-bech32-bitcoin and prepare a PR to fix any emerging incompatibilities. Maybe this will show me some rough edges which I can improve.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -32,7 +32,7 @@
 //! ```rust
 //! use bech32::Bech32;
 //!
-//! let b = Bech32::new("bech32".into(), vec![0x00, 0x01, 0x02]).unwrap();
+//! let b = Bech32::new_check_data("bech32".into(), vec![0x00, 0x01, 0x02]).unwrap();
 //! let encoded = b.to_string();
 //! assert_eq!(encoded, "bech321qpz4nc4pe".to_string());
 //!
@@ -51,7 +51,7 @@ use std::str::FromStr;
 use std::fmt::{Display, Formatter};
 
 /// Integer in the range `0..32`
-#[derive(PartialEq, Debug, Copy, Clone)]
+#[derive(PartialEq, Debug, Copy, Clone, Default)]
 #[allow(non_camel_case_types)]
 pub struct u5(u8);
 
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -50,41 +50,140 @@ use std::{error, fmt};
 use std::str::FromStr;
 use std::fmt::{Display, Formatter};
 
+#[cfg(feature = "try_from")]
```
clarkmoody: What is the compatibility of `try_from`? Other changes were made to this library to push compatibility back to 1.14.

sgeisler: AFAIK it's new in 1.25, that's why I put it behind a feature gate. You can use it if your rustc version is new enough, but you can ignore it otherwise. Unfortunately I couldn't find a way to express "only activate if rustc version >= 1.25.0", so the user has to activate it manually. But this reminds me of a TODO: run the stable test with the feature activated and add a test for `TryFrom`.

src/lib.rs
```diff
@@ -50,41 +50,140 @@ use std::{error, fmt};
...
+use std::convert::TryFrom;
+
+/// Integer in the range `0..32`
+#[derive(PartialEq, Debug, Copy, Clone, Default)]
+#[allow(non_camel_case_types)]
+#[repr(C)]
```
sgeisler: remove `repr(C)`, was used to do test with `std::mem::transmute`

clarkmoody:
You've got a few `TODO`s still in the code. Were you doing to address those before moving forward with this?

sgeisler:
I will defineatly address them, that's why I marked the PR as WIP. But right now I'm using the API with bech32-bitcoin and lightning-invoice to find rough edges. I already found some and will fix them next.

src/lib.rs
sgeisler: #33417

apoelstra:
Can you squash down the TODO fixes? Note that you can selectively squash with `git rebase -i`.

apoelstra:
It's very hard to review changes where commits introduce things that other commits remove.

sgeisler:
With some squashing and rebasing I cut the commit count down to four. I hope this is a better starting point for a review.

clarkmoody:
I will try to review soon.

sgeisler:
I just saw that I didn't increase the version yet. I guess it would be 0.4.0?

clarkmoody:
👍 I was thinking about that earlier. Thanks

PR #23, opened 2018-06-22
Fix missing range checks

### LINKED ISSUE #22: Panic when decoding bech32 string containing an uppercase 'O'
Fuzz tests in rust-bitcoin/rust-bitcoin#100 uncovered the bug that if the bech32 data part contains an uppercase 'O' the parser panics. This is due to insufficient checks that only check the lowercase variants of the forbidden characters (before normalizing them to their lowercase variant):

```rust
if b == b'1' || b == b'b' || b == b'i' || b == b'o' {
    return Err(Error::InvalidChar(b))
}
```

To avoid such issues in the future I will have a look at fuzz testing this crate (#21).

sgeisler:
Fixes #22. 

## TODO:

- [x] rewrite `from_str_lenient` to catch all bad characters
- [x] adapt tests that use `Error::InvalidChar` since its signature was updated
- [x] validate that all bad characters are `-1` in `CHARSET_REV`
- [x] write tests to show the bug is fixed
- [x] have a look at the checks for mixed case, they seem overly complicated

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -551,6 +551,8 @@ mod tests {
                 Error::InvalidLength),
             ("x1b4n0q5v",
                 Error::InvalidChar('b')),
+            ("ABC1DEFGOH",
+                Error::InvalidChar('O')),
             ("li1dgmt3",
                 Error::InvalidLength),
             ("de1lg7wt\u{ff}",
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -63,8 +63,9 @@
 #![deny(unused_mut)]
 
 use std::{error, fmt};
-use std::str::FromStr;
+use std::ascii::AsciiExt;
 use std::fmt::{Display, Formatter};
+use std::str::FromStr;
 
 /// Integer in the range `0..32`
 #[derive(PartialEq, Eq, Debug, Copy, Clone, Default, PartialOrd, Ord, Hash)]
@@ -249,9 +250,9 @@ impl Bech32 {
                 return Err(Error::InvalidChar(c))
             }
 
-            if c.is_ascii_lowercase() {
+            if c.is_lowercase() {
                 has_lower = true;
-            } else if c.is_ascii_uppercase() {
+            } else if c.is_uppercase() {
                 has_upper = true;
             }
 
@@ -540,8 +541,8 @@ mod tests {
         let pairs: Vec<(&str, Error)> = vec!(
             (" 1nwldj5",
                 Error::InvalidChar(' ')),
-            ("\x7f1axkwrx",
-                Error::InvalidChar(0x7f as char)),
+            ("abc1\u{2192}axkwrx",
+                Error::InvalidChar('\u{2192}')),
             ("an84characterslonghumanreadablepartthatcontainsthenumber1andtheexcludedcharactersbio1569pvx",
                 Error::InvalidLength),
             ("pzry9x0s0muk",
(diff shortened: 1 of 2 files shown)


sgeisler:
The tests fail due to rustup issues with nightly: rust-lang/rust#51699.

src/lib.rs
```diff
@@ -242,34 +243,28 @@ impl Bech32 {
...
-            if b >= b'a' && b <= b'z' {
+
+            if c.is_lowercase() {
                 has_lower = true;
+            } else if c.is_uppercase() {
+                has_upper = true;
             }
 
-            // Uppercase
-            let c = if b >= b'A' && b <= b'Z' {
```
tamasblummer: was dropping conversion to lower case intentional?

clarkmoody: I like this use of `CHARSET_REV` in testing for proper range.

src/lib.rs
```diff
@@ -242,34 +243,28 @@ impl Bech32 {
...
+        let mut data_bytes = raw_data.chars().map(|c| {
+            // Only check if c is in the ASCII range, all invalid ASCII characters have the value -1
+            // in CHARSET_REV (which covers the whole ASCII range) and will be filtered out later.
+            if !c.is_ascii() {
+                return Err(Error::InvalidChar(c))
             }
-            // Lowercase
-            if b >= b'a' && b <= b'z' {
+
+            if c.is_lowercase() {
```
sgeisler: The `is_lowercase` and `is_uppercase` functions are probably slower (since they work for all Unicode characters) than the simple range checks used before, but they seem so much more idiomatic. The ascii-equivalent (`is_ascii_lowercase`) is still `nightly`-only, so if we wanted to avoid handwritten range checks we had to use our own trait for that. But I don't see the need for such optimizations right now.

src/lib.rs
```diff
@@ -242,34 +243,28 @@ impl Bech32 {
...
                 has_lower = true;
```
sgeisler: I didn't find a better (and preferably shorter/simpler) way of doing the checks for "all characters have the same case". At least none that doesn't need some bigger refactoring of the HRP processing, so I will leave it for now and maybe open another PR dedicated to refactoring.

sgeisler:
Since this seems to be only an internal bug fix I increased the version from `0.4.1` to `0.4.2`. Can anyone review the changes, please? @clarkmoody @tamasblummer

src/lib.rs
```diff
@@ -402,7 +397,7 @@ pub enum Error {
     /// The data or human-readable part is too long or too short
     InvalidLength,
     /// Some part of the string contains an invalid character
-    InvalidChar(u8),
+    InvalidChar(char),
```
clarkmoody: Is this enum variant change a breaking change that needs a major version bump?

src/lib.rs
clarkmoody: The implementation of `is_lowercase` is:
```rust
    pub fn is_lowercase(self) -> bool {
        match self {
            'a'...'z' => true,
            c if c > '\x7f' => derived_property::Lowercase(c),
            _ => false,
        }
    }
```
So for ASCII characters, it should be fairly performant. `is_uppercase` has a similar structure.

src/lib.rs
sgeisler: Good catch, sorry, I will have to bump the version to `0.5.0`. Would you generally agree that it's better to process unicode strings char-wise instead of their UTF-8 representation byte-wise? Because this change is what made this API break necessary.

clarkmoody:
Pushed `0.5.0` to crates.io

PR #24, opened 2018-07-05
Remove unused AsciiExt Import

clarkmoody:
The `AsciiExt` import was causing lots of warnings and was unused.


clarkmoody:
@sgeisler I see the Rust 1.14 build fail.

Is there a way to suppress these warnings of the deprecated package?

PR #26, opened 2018-07-05
Treat warnings as errors when running tests on travis CI

sgeisler:
To avoid regressions like #24 I'd like to treat all warnings like errors when running our tests on travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,7 @@ rust:
   - nightly
   - 1.14.0  # shipping debian version 2018 03
 cache: cargo
+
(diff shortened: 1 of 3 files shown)


sgeisler:
Since this PR doesn't contain any API/functionality changes or bug fixes it shouldn't get it's own release on crates.io. I'd say we can just merge it. Nobody besides us should care about this change because it only affects travis builds (and it will work as soon as it's on master). Just my 2satoshi, if you really want it or have some reason I didn't think of I will of course bump the version.

PR #27, opened 2018-11-26
Add Fuzzing

### LINKED ISSUE #21: Implement fuzz testing
Fuzz testing in `rust-bitcoin` found an error in `rust-bech32` (rust-bitcoin/rust-bitcoin#100 ). This crate should have its own fuzz testing.

### LINKED ISSUE #28: Decoding our own encoded bech32 strings can fail
Found this while working on fuzz testing (#27 ). 

Cause: we don't check for forbidden characters in the `hrp` when constructing a `Bech32` struct.

sgeisler:
Add fuzz tests to rust-bech32, fixes #21, fixes #28 . Not ready for review yet.

- [x] decode random data
- [x] encode/decode random data
- [ ] ~decode random encoded data from bitcoin core implementation~
- [ ] ~encode random data and test if core can parse it~
- [x] travis integration and cleanup
- [x] fix #28 to make fuzz tests not fail


clarkmoody:
👍 Thanks for working on this.

sgeisler:
Since it's better to have some fuzzing than no fuzzing I'd like to get this merged without the core cross-fuzzing. At some point I will probably revisit this to implement cross-implementation fuzzing.

This PR also includes fixes for #28 which change the API. So if this is merged the version should be bumped. Depending on your preferences I could just add one commit doing that to this PR.

clarkmoody:
Please add a commit to bump version.

I will dig in and review.

src/lib.rs
```diff
@@ -172,11 +176,28 @@ impl<T: AsRef<[u8]>> ToBase32<Vec<u5>> for T {
...
 impl Bech32 {
     /// Constructs a `Bech32` struct if the result can be encoded as a bech32 string.
+    ///
+    /// # Errors
+    /// * **MixedCase**: if the HRP contains upper case letters. Encoders must use lower case.
+    /// * **InvalidChar**: if the HRP contains any non-ASCII characters (outside 33..=126)
+    /// * **InvalidLength**: if the HRP is outside 1..83 characters long
+    ///
+    /// # Deviations from standard
+    /// * No length limits are enforced for the data part
```
sgeisler: It was much simpler to not enforce this invariant because lightning uses much longer than standard bech32 invoices. But because we need to enforce the same invariants when parsing and constructing that would require two structs `Bech32` and `LenientBech32`. I think it's not worth the added complexity.

clarkmoody: Agreed.

PR #30, opened 2019-04-22
Add `encode_to_fmt` method for efficient encoding

stevenroose:
Export a public method that directly writes a `Vec<u5>` to an `fmt::Formatter`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -226,19 +226,26 @@ impl Bech32 {
     }
 }
 
+/// Encode a bech32 payload to an [fmt::Formatter].
+pub fn encode_to_fmt<T: AsRef<[u5]>>(fmt: &mut fmt::Formatter, hrp: &str, data: T) -> fmt::Result {
+    let hrp_bytes: &[u8] = hrp.as_bytes();
+    let checksum = create_checksum(hrp_bytes, data.as_ref());
+    let data_part = data.as_ref().iter().chain(checksum.iter());
+
+    write!(
+        fmt,
+        "{}{}{}",
+        hrp,
+        SEP,
+        data_part
+            .map(|p| CHARSET[*p.as_ref() as usize])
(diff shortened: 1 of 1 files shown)


stevenroose:
@sgeisler would you like an additional method directly producing a string based on this? I don't need it, but it makes sense because using `fmt::Formatter` is kinda `Display`-only.

```rust
 pub fn encode<T: AsRef<[u5]>>(hrp: &str, data: T) -> String;
```

stevenroose:
So you are saying we should make sure that the `hrp` part is also written in the same case as the data? That makes sense.

sgeisler:
Yes, it's part of the specification that a bech32 string has only one case. And if encoding only lowercase is allowed and it's left to the application to transform it to uppercase if QR encoding is intended.

And I think these invariants should be part of a struct's "contract". That's why a `(&str, &[u8])` tuple is not sufficient in my opinion.

stevenroose:
> Yes, it's part of the specification that a bech32 string has only one case. And if encoding only lowercase is allowed and it's left to the application to transform it to uppercase if QR encoding is intended.

The check that makes sure the prefix is valid is quite negligible, no? Most prefixes will be < 10 characters long and the check is just a byte comparison IIRC. The only question is then whether to panic or Err on an incorrect one. `encode_to_fmt` is kinda made to return an `fmt::Result`. A `Result<fmt::Result, bech32::Error>` would be a bit awkward, but I guess then we could unwrap it in the `Display impl.

stevenroose:
Superseded by https://github.com/rust-bitcoin/rust-bech32/pull/34. Let's move discussion there.

dongcarl:
@stevenroose Okay to close?

sgeisler:
I think #35 included #34 which came out of this discussion, and got merged, so this can be closed too.

PR #31, opened 2019-04-23
Add `decode` method for raw decoding

stevenroose:
Export public method to decode into the raw HRP and the data bytes.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -242,58 +242,59 @@ impl Display for Bech32 {
     }
 }
 
-impl FromStr for Bech32 {
-    type Err = Error;
+/// Decode a bech32 string into the raw HRP and the data bytes.
+/// The HRP is returned as it was found in the original string,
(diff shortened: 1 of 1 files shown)


stevenroose:
One quick piece of feedback: do you think it's worth allocating a `String` to return the HRP and case it to lowercase instead of returning the original HRP? We kinda already do that with the `hrp_bytes` that is used for the checksum validation.

sgeisler:
It's probably easier for potential users of this function if it always outputs a lowercase HRP. It could return a `Cow<'a, str>` for that purpose. But maybe we want to generally rethink the ownership of `Bech32` as I wrote in #30? In that case the function could return a minimally owning bech32 struct.

stevenroose:
Superseded by https://github.com/rust-bitcoin/rust-bech32/pull/34. Let's move discussion there.

sgeisler:
Included in #35

PR #32, opened 2019-04-23
Remove redundant range check

sgeisler:
Since `u5` is defined in this crate it can be instantiated directly avoiding a redundant range check.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -309,7 +309,7 @@ impl FromStr for Bech32 {
                 return Err(Error::InvalidChar(c));
             }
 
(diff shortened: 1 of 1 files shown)


clarkmoody:
Can we close this after the work done on #35?

sgeisler:
I thought about that too. It isn't redundant and would just need a rebase, but while it's obvious to us *now* that the invariants hold it might not be in the future, so maybe leaving it as is is actually best.

I'll close it for now, if you disagree feel free to reopen, I should have put more thought into it before opening the PR (just noticed it when reviewing something else).

PR #35, opened 2019-06-16
Allocationless ecoding

sgeisler:
Builds on top of and closes #34 and adds a `Bech32Writer` struct that allows allocationless encoding to an underlying `std::fmt::Writer`.

- [x] ~fuzz `[u8]` -> `[u5]` encoding~ is covered by current (adapted) fuzz tests
- [x] remove redundant encoding code
- [x] fix rust 1.22 compatibility

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -30,29 +30,14 @@
 //! # Examples
 //!
 //! ```rust
-//! use bech32::Bech32;
+//! use bech32::{self, FromBase32, ToBase32};
 //!
-//! let b = Bech32::new_check_data("bech32".into(), vec![0x00, 0x01, 0x02]).unwrap();
-//! let encoded = b.to_string();
-//! assert_eq!(encoded, "bech321qpz4nc4pe".to_string());
+//! let encoded = bech32::encode("bech32", vec![0x00, 0x01, 0x02].to_base32()).unwrap();
+//! assert_eq!(encoded, "bech321qqqsyrhqy2a".to_string());
 //!
-//! let c = encoded.parse::<Bech32>();
-//! assert_eq!(b, c.unwrap());
-//! ```
-//!
-//! If the data is already range-checked the `Bech32::new` function can be used which will never
-//! return `Err(Error::InvalidData)`.
-//!
-//! ```rust
-//! use bech32::{Bech32, u5, ToBase32};
-//!
-//! // converts base256 data to base32 and adds padding if needed
-//! let checked_data: Vec<u5> = [0xb4, 0xff, 0xa5].to_base32();
-//!
-//! let b = Bech32::new("bech32".into(), checked_data).expect("hrp is not empty");
-//! let encoded = b.to_string();
-//!
-//! assert_eq!(encoded, "bech321knl623tk6v7".to_string());
+//! let (hrp, data) = bech32::decode(&encoded).unwrap();
+//! assert_eq!(hrp, "bech32");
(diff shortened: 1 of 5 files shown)


stevenroose:
I tested this briefly with https://github.com/rust-bitcoin/rust-bitcoin/pull/255 and it seems to work.

stevenroose:
There is something not working in the Rust 1.14 test. Could you look at that? Then we can get this merged and https://github.com/rust-bitcoin/rust-bitcoin/pull/255 as well.

sgeisler:
Didn't we agree on 1.22 as oldest supported rust Version? That's why I
changed the Travis file. Of course we could still try to keep our
compatibility, but is it really worth the extra effort?

stevenroose:
Oh yeah no sure 1.22 is fine, it's rust-bitcoin's version. Ah I clicked the old Travis report, my bad. Sure then it's good to go IMO. I gave my approval, but I'm not an official contributor to this repo. Can you merge without ACKs?

sgeisler:
No, I'd also like to hear @clarkmoody's opinion first.

clarkmoody:
Looking over this now. Quite a lot of changes 😄 

Can we update the examples in the `README` to incorporate these changes?

src/lib.rs
```diff
@@ -384,22 +519,18 @@ const SEP: char = '1';
 
 /// Encoding character set. Maps data value -> char
 const CHARSET: [char; 32] = [
-    'q','p','z','r','y','9','x','8',
-    'g','f','2','t','v','d','w','0',
-    's','3','j','n','5','4','k','h',
-    'c','e','6','m','u','a','7','l'
+    'q', 'p', 'z', 'r', 'y', '9', 'x', '8', 'g', 'f', '2', 't', 'v', 'd', 'w', '0', 's', '3', 'j',
```
clarkmoody: It looks like these have been auto-formatted, which makes them a little harder to visually verify against the BIP-0173 spec and other implementations. Not a deal-breaker, since errors will show up in test, but if there's a clean way to skip auto-formatting / linting for these few lines (and for `CHARSET_REV` below), I would prefer that.

sgeisler: I reverted the changes and added an ignore attribute.

src/lib.rs
```diff
@@ -163,143 +230,232 @@ impl FromBase32 for Vec<u8> {
...
+///
+/// Returns the HRP in lowercase..
+pub fn decode(s: &str) -> Result<(String, Vec<u5>), Error> {
+    // Ensure overall length is within bounds
+    if s.len() < 8 {
+        return Err(Error::InvalidLength);
+    }
+
+    // Split at separator and check for two pieces
+    let (raw_hrp, raw_data) = match s.rfind("1") {
```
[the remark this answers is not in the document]
sgeisler: Fixed. It should be more efficient too (ignoring compiler optimization) since it's now matching against a `char` instead of a `&str`.

src/lib.rs
clarkmoody: Looks like tests are failing from that new attribute.

sgeisler: Yeah, I'm already searching for the right attribute to deactivate that, I used it recently but can't remember, I guess it's `#[allow(unknown_attributes)]`.

sgeisler: I ended up just removing the attributes. `unknown_attributes` requires a feature only available on nightly.

PR #41, opened 2019-12-31
Convert HRP to lowercase for encoding purposes

### LINKED ISSUE #40: bech32::encode computes incorrect checksum for uppercase HRPs
[BIP 173](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki) states:

> **Uppercase/lowercase**
>
> The lowercase form is used when determining a character's value for checksum purposes.
>
> Encoders MUST always output an all lower-case Bech32 string.

`bech32::encode` verifies the provided HRP is internally consistent, but then ignores the returned case type and uses the HRP directly:

https://github.com/rust-bitcoin/rust-bech32/blob/16bbf7e011ae12d5c434dc57dc7471705e855fbe/src/lib.rs#L384-L385

This causes two issues:

- The encoded string is in mixed-case, which is likely what the caller intended but is invalid per the spec. Either the HRP should be auto-lower-cased, or an error should be returned.
- The encoded string is invalid and cannot be decoded, because the decoder (correctly) lower-cases the HRP before verifying the checksum.

clarkmoody:
Fixes #40

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -53,6 +53,7 @@
 #![cfg_attr(feature = "strict", deny(warnings))]
 
 use std::{error, fmt};
+use std::borrow::Cow;
 
 // AsciiExt is needed for Rust 1.14 but not for newer versions
 #[allow(unused_imports, deprecated)]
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -381,14 +383,18 @@ pub fn encode_to_fmt<T: AsRef<[u5]>>(
     hrp: &str,
     data: T,
 ) -> Result<fmt::Result, Error> {
-    check_hrp(&hrp)?;
-    match Bech32Writer::new(hrp, fmt) {
+    let hrp_lower = match check_hrp(&hrp)? {
+        Case::Upper => hrp.to_lowercase(),
+        Case::Lower | Case::None => String::from(hrp),
```
sgeisler: The string allocation could be avoided by using `Cow<str>`. 

```suggestion-2
        Case::Upper => Cow::Owned(hrp.to_lowercase()),
        Case::Lower | Case::None => Cow::Borrowed(hrp),
```

clarkmoody: Great point. I'll make that modification.

clarkmoody: Added in d2e4c49.

PR #42, opened 2020-01-01
Improved docs, fixed clippy warnings, fixed cargo fmt

duesee:
When reading the code I made some minor improvements I wanted to share. Notably, I tried to make the usecase for Bech32 more clear and fix all clippy warnings and "incorrect" formatting.

Have you thought about introducing a CI (e.g., via GitHub actions) which checks against cargo audit, cargo clippy and cargo fmt?


clarkmoody:
Interesting. I'll circle back after #41 is merged. We'll probably need a rebase then as well.

sgeisler:
The changes look good, some transformations from index to iterator loops might even make the code faster.

The only problem I don't have a good solution for right now is how to handle the `rustfmt::skip` attribute in older rust versions (before macros were scoped). Currently it makes our tests fail for rust `1.22`. Maybe we can make the attribute conditional and only run `rustfmt` with more recent rust versions. But I fear the error occurs when parsing the AST, so even disabling the code with `cfg` won't help. What we could do in that case is removing the attribute using `sed` when running the `1.22` test since it isn't needed there anyway, but that's an ugly hack.

If we accept `rustfmt` changes we should definitely add a `rustfmt` check to our [already existing CI pipeline](https://github.com/rust-bitcoin/rust-bech32/blob/master/.travis.yml) otherwise non-conforming code will be checked in again. Clippy is a bit of a problem since not all fixes are compatible with older rust versions. `cargo audit` seems most useful as a cron job and not for every PR, we could activate weekly builds in travis and only run it on master.

clarkmoody:
Needs rebase.

duesee:
Is there a reason to support Rust 1.22? Maybe it could be `edition = '2018'`? :-)

sgeisler:
The rationale for older rust versions is given in rust-bitcoin/rust-bitcoin#338, although I can't remember why it's exactly 1.22. 1.24 seems to be the version we really want to support, but I fear that won't help with our problem.

duesee:
Hm... okay, so I don't see any easy way of supporting rustfmt's skip. `#[rustfmt::skip]` does not work in 1.22.0, Clippy lints on `#[cfg_attr(rustfmt, rustfmt_skip)]` and I think also don't think that `#[warn(clippy::deprecated_cfg_attr)]` will work in 1.22 either nor should it be used. I just bypassed the formatting so that not to generate too much hassle with this PR.

duesee:
Regarding cargo audit: having it run on every PR prevents pulling in already vulnerable dependencies -- who knows? :-)

duesee:
I would be glad if you could do it :-)

.travis.yml
```diff
@@ -11,6 +11,14 @@ script:
 
 jobs:
   include:
+  - stage: best practices
+    rust: stable
+    install:
+      - rustup component add rustfmt
+      - rustup component add clippy
+    script:
+      - rustfmt --check src/lib.rs
```
sgeisler: Is there a better way to do this? Apparently `cargo fmt` doesn't support the `--check` flag, so I have to use `rustfmt` directly and specify the files to check.

sgeisler:
Oh, I just noticed that there are merge conflicts now. I'll try to rebase.

duesee:
Thank you @sgeisler! The CI should catch everything now. Having cargo audit run on every push might still be a good (and easy to implement) idea. But it could be added at a later time and issue.

sgeisler:
#response_container_BBPPID{font-family: initial; font-size:initial; color: initial;}  Oh yeah, I totally forgot about that. Will amend the last commit tomorrow.                                                                                                                                                                                                                                                                                                                        From: notifications@github.comSent: January 11, 2020 23:46To: rust-bech32@noreply.github.comReply-to: reply@reply.github.comCc: geisler.sebastian@googlemail.com; mention@noreply.github.comSubject: Re: [rust-bitcoin/rust-bech32] Improved docs, fixed clippy warnings, fixed cargo fmt (#42)  Thank you @sgeisler! The CI should catch everything now. Having cargo audit run on every push might still be a good (and easy to implement) idea. But it could be added at a later time and issue.

—You are receiving this because you were mentioned.Reply to this email directly, view it on GitHub, or unsubscribe.

sgeisler:
I though about `cargo audit` again and it doesn't make much sense for us anyway since we don't have any dependencies.

duesee:
You don't have dependencies *now* ;-)

clarkmoody:
I think it should be a long-term goal that this library has no dependencies...

PR #44, opened 2020-03-11
Convert to Edition 2018 --- contains #43

Ericson2314:
#43 is not strictly needed, but it's nice to remove the `extern crate` I had to add to that one in this.


sgeisler:
Nack for now, should be included once the minimum rust version gets bumped again.

Ericson2314:
The preference/decision (I'm not sure which) in #43 was to bump the MSRV, after which there is no reason not to do this.

tcharding:
We can close this one, edition 2018 bump was done in https://github.com/rust-bitcoin/rust-bech32/pull/57

PR #45, opened 2020-03-12
Expand the amount of core-only

Ericson2314:


Ericson2314:
This fails because of missing `impl`s with rust 1.22. How important is rust 1.22?

src/lib.rs
```diff
@@ -308,22 +354,37 @@ impl<T: AsRef<[u8]>> Base32Len for T {
...
+}
+
+impl<'f, U: AsRef<[u8]>> CheckBase32<()> for U {
+    type Err = Error;
+
+    fn check_base32(self) -> Result<(), Error> {
+        self.as_ref()
+            .iter()
+            .map(|x| u5::try_from_u8(*x).map(|_| ()))
+            .collect::<Result<(), Error>>()
```
sgeisler: This should do the same imo:

```suggestion
            .find(|r| r.is_err())
            .unwrap_or(Ok(()))
```

Ericson2314: Oh, awesome it's that simple!

reuvenpo:
If you set the minimum version of rust to 1.36 you can use the `alloc` crate with `no_std` and make only minimal changes to the crate, because it supports `Vec`s and other collection types that are used in this crate :)

Ericson2314:
@reuvenpo I want to use this on embedded systems like hardware wallets where dynamic allocation is highly discouraged.

reuvenpo:
>I want to use this on embedded systems like hardware wallets where dynamic allocation is highly discouraged
That's reasonable, if it were solved well this way we wouldn't mind using it in my own project :)

>There is no reason we couldn't do both approaches
If it makes sense in the code, sure. My use case required just using bech32 in an environment that has alloc but not std, so i just whipped up #46 (which was the fastest way to achieve that) and we're using that for now. We won't push for this duality if it's a hassle, cause your solution would work well too.

Ericson2314:
I adapted this in light of the existing `std` feature.

tcharding:
Hi @Ericson2314, I just put up a PR that aims to do more or less the same thing as this one, I did it without going back through open PRs so I missed this one. I don't want to seem like I was ignoring this so pinging you now. If you want to pick this work up I can close mine since you were first.

Ericson2314:
@tcharding no worries, I haven't touched this stuff in years. Mine does more things than yours, generalizing some interfaces, so why don't we just do yours first and then mine as follow-up?

Ericson2314:
@tcharding I rebased this on top as described, and polished it a bit more.

tcharding:
Hey @Ericson2314, I'm not exactly sure what this PR is trying to achieve. "allow doing more things without alloc" doesn't give me quite enough context to evaluate the PR, I like the additional embedded test code but the base N stuff seems orthogonal, or am I missing something? Perhaps you could write a PR description justifying the proposed changes and if I'm correct in thinking the base N stuff is unrelated to the embedded test improvements perhaps split it out into a separate commit. Thanks

Ericson2314:
To be honest I forget all the details, but see the tests I've enabled, and the new ones I've written.

If your PR is merged first, it will be easier both for you all to read the code, and me to write the expanded description!

tcharding:
Hey @Ericson2314, other PRs merged now so if you are still keen to work on this the path is clear.

src/lib.rs
```diff
@@ -108,25 +111,40 @@ impl AsRef<u8> for u5 {
     fn as_ref(&self) -> &u8 { &self.0 }
 }
 
-/// Interface to write `u5`s into a sink
-pub trait WriteBase32 {
+/// Interface to write `u(2*n)`s into a sink
+pub trait WriteBaseN<T>
```
apoelstra, line 115: In be4af63082fdbe52c7de6cb5186c2f51be6a7d0b:

I find this rename very confusing. It looks like it's generic over the number base, but the `write` implementation still calls `write_u5` ... which now no longer has anything to do with u5s.

I think we should leave this trait alone and make `WriteBase256` completely independent (and probably it does not need the single `write_u5` method since Rust has decent first-class support for byteslices).

Ericson2314: Oops not renaming `write_u5` was just an oversight.

I kept it one trait to I can have the polymorphic `Vec` and `ArrayVec` implementations, but I suppose I could just copy paste them for each trait.

apoelstra:
Looks like we require `cargo clippy` to pass. Just minor fixes needed.

src/lib.rs
apoelstra: I would suggest copy/pasting because I think `u8` serialization is "just dumping bytes" and `u5` serialization as a nontrivial conversion, so they feel like distinct things to me even though the code is pretty-much the same.

Ericson2314: OK

Ericson2314:
OK fixed conflicts, and fixed clippy.

apoelstra:
You might need to use a more recent version of clippy :).

apoelstra:
@Ericson2314 I'd like to get #88 merged, which will probably cause you some rebasing work. Are you still actively working on this, and do you want us to hold off?

Ericson2314:
Sorry, I'll try to push on this a bit more tonight.

Ericson2314:
Tests pass, now just getting rid of `WriteBaseN` as requested.

Ericson2314:
That is done too. Should be all ready.

tcharding:
Is the point of this PR to enable encoding a slice of `u5`s to bech32 without an allocator? If so I think this would be better solved by using iterators. The `arrayvec` dependency breaks MSRV so I don't see how we can use it. (I tried it out with Rust 1.48 as well.)

An update to the PR description and title would be super helpful please - took me a while to work out what was going on.

apoelstra:
The arrayvec msrv looks like it's 1.51, which is not crazy far from our MSRV, and it's an optional dep which looks independent of the other alloc stuff. I think it'd be okay to say "if you enable this you con't use our MSRV".

apoelstra:
So, I'd be ready to ACK this PR, but I find @tcharding idea of replacing these `Write` traits with an iterator (which maintains only a small amount of state, and a pointer to its input) intriguing. @Ericson2314 does this seem plausible to you?

I'm okay with keeping an optional `arrayvec` dep but if we could avoid it that'd be even better.

tcharding:
To be totally transparent I got this idea from https://github.com/rust-bitcoin/rust-bech32/issues/60 (posted by @Kixunil) but I have not worked out _exactly_ what it means to implement it. Am happy to learn more about iterators to do it but equally happy to see it done and learn by reviewing it :)

src/lib.rs
```diff
@@ -262,18 +282,71 @@ pub trait FromBase32: Sized {
...
+
+#[cfg(feature = "arrayvec")]
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Debug)]
+/// Combination of Errors for use with array vec
+pub enum ComboError {
+    /// Error from this crate
+    Bech32Error(Error),
+    /// Error from `arrayvec`.
+    CapacityError(CapacityError),
+}
```
Kixunil, line 330: I think `WriteError` would be a clearer name.

src/lib.rs
```diff
@@ -367,28 +437,42 @@ impl<T: AsRef<[u8]> + ?Sized> Base32Len for T {
...
+
+impl<U: AsRef<[u8]>> CheckBase32<()> for U {
+    type Err = Error;
+
+    fn check_base32(self) -> Result<(), Error> {
+        self.as_ref()
+            .iter()
+            .map(|x| u5::try_from(*x).map(|_| ()).map_err(Error::TryFrom))
+            .find(|r| r.is_err())
+            .unwrap_or(Ok(()))
```
Kixunil, line 471: This looks like a really convoluted way of checking validity. Where is this needed?

Kixunil (changes requested):

src/lib.rs
```diff
@@ -162,7 +163,26 @@ pub trait WriteBase32 {
...
+    /// Write a `u8` slice
+    fn write(&mut self, data: &[u8]) -> Result<(), Self::Err> {
+        for b in data {
+            self.write_u8(*b)?;
+        }
+        Ok(())
+    }
+
+    /// Write a single `u8`
+    fn write_u8(&mut self, data: u8) -> Result<(), Self::Err> { self.write(&[data]) }
```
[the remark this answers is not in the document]
apoelstra: I think it's fine, it's a pretty common idiom for a trait to have infinite recursion by default, so that you can implement either one of the methods. If you try to run it, it will crash immediately (actually, I think the compiler will even flag it).

embedded/no-allocator/Cargo.toml
```diff
@@ -5,11 +5,24 @@ readme = "README.md"
...
 panic = "abort"
 
 [profile.release]
 panic = "abort"
-
-[dependencies]
-bech32 = { path = "../../", default_features = false }
+codegen-units = 1 # better optimizations
+debug = true # symbols are nice and they don't increase the size on Flash
+lto = true # better optimizations
```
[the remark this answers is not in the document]
Ericson2314: I think I was just copying what the other test did (or used to do) to try to go with the flow.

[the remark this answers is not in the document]
Kixunil: Maybe they are intended for benches? Those should have `[profile.release]` above these settings.

Ericson2314:
Would it be possible to merge this and then take a stab at the iterators approach? Would like to do it, but I am worried about finding the time for another "creative" investigation, vs the more tasks that are the other review items. I agree it does sound like a good idea.

apoelstra:
@Ericson2314 for sure, let's just merge this for now. I don't think we have a release of this crate coming up anytime soon, and this PR is blocking some other things.

But could you first fix the following minor comments from Kix?:

* remove the new `lto = true` stuff from embedded/no-allocator/Cargo.toml
* rename `CapacityError` to `WriteError`

Then later we can explore the iterator thing and revisit the `check_base32` API.

Ericson2314:
No problem @apoelstra, happy to do all those things. Thank you for agreeing to defer the iterator exploration.

Ericson2314:
I renamed `CapacityError` as requested, as to the `lto` stuff. C.f. `with-allocator`. I tried to make it more like before (not just the `lto` thing but in general), but that didn't work so well when the goal is to actually *run* the no-std binary.

Right now, `with-allocator` and `no-allocator` should be set up exactly the same, and I recommend if we do get rid of the `lto` thing for one of them, we should get rid of it for both of them.

apoelstra:
Lol, CI is failing, something to do with LTO ..... maybe there **was** a reason for those flags.

tcharding (approved):
ACK 4b0c640727c1a7158206a225f8ac20b486a4b949

Ericson2314:
Yeah CI wasn't actually running the thing as I intended before the previous push.

We'll see if I fixed it. Also added a new readme for both weird ARM tests.

Ericson2314:
Yay OK the last commit did fix everything!

tcharding (approved):
ACK 42b34469027781d97e5c1d3107b25dee68acfe65

tcharding:
Thanks for sticking with it man.

apoelstra:
This is super cool! I didn't know how to use qemu in CI before. We should use this to get a BE system to test rust-secp against.

apoelstra (approved):
ACK 42b34469027781d97e5c1d3107b25dee68acfe65

embedded/no-allocator/memory.x
```diff
@@ -0,0 +1 @@
+../with-allocator/memory.x
```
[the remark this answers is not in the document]
Ericson2314: OK I'll just copy the file then no problem.

apoelstra (approved):
ACK 32c87daa3072be1a83717fe789a34556442c3553

apoelstra:
Hooray! That time it was accepted.

Thanks for your contribution @Ericson2314 and glad to have you poking at these projects!

Ericson2314:
And thanks everyone too for helping me get this over the finish line!

tcharding:
Nice one man, thank you.

PR #50, opened 2021-02-12
Bech32m Support

### LINKED ISSUE #48: Support bech32m

clarkmoody:
Adds support for [BIP-0350](https://github.com/bitcoin/bips/blob/master/bip-0350.mediawiki), Bech32m encoding. Bumps version to `0.8.0`.

### Breaking Changes

- `decode` returns the `variant` as third value upon success. This will be used downstream when validating SegWit addresses against the allowed encoding variant.
- `encode`, `encode_to_fmt`, and `Bech32Writer::new`  require `variant` parameters. When creating SegWit addresses, the proper variant will be selected based on the script version.

Closes #48


src/lib.rs
```diff
@@ -400,22 +410,52 @@ pub fn encode_to_fmt<T: AsRef<[u5]>>(
     }
 }
 
+/// Used for encode/decode operations for the two variants of Bech32
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
+pub enum Variant {
+    /// The original Bech32 described in [BIP-0173](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki)
+    Bech32,
```
dr-orlovsky, line 417: can we pls have the actual constants assigned as the values here? When we will not need `BECH32_CONST`, `BECH32M_CONST`,  and can simply use `Variant::Bech32 as u32` instead of `constant()` function call (and avoid one call)

src/lib.rs
```diff
@@ -400,22 +410,52 @@ pub fn encode_to_fmt<T: AsRef<[u5]>>(
     }
 }
 
+/// Used for encode/decode operations for the two variants of Bech32
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
+pub enum Variant {
```
dr-orlovsky, line 415: I suggest also to add `repr(u32)` here

dr-orlovsky (commented):
Have not reviewed against the spec; few suggestion after the first code read

src/lib.rs
sgeisler: I'm fine with @clarkmoody's version, it's more "standard" imo, do you feel strongly about this?

sgeisler (approved):
I like it, very elegant :smiley: 

ACK 411d8c45f0637da3c5a04c516f405f96113d60c7

sgeisler:
Btw, do you know what's wrong with CI? It just doesn't run that one job, maybe we should switch to GH actions too …

src/lib.rs
clarkmoody: I'm not exactly sure what you mean @dr-orlovsky. Could you post a code example?

Performance-wise, the compiler will inline these simple methods for sure.

clarkmoody:
> Btw, do you know what's wrong with CI? It just doesn't run that one job, maybe we should switch to GH actions too …

I think it's a modern version of Clippy failing on our 1.22-compatible code.

I agree that we should switch to GH actions. I'll work on that. Do you think it's also time to raise our version to 1.29 to match the `rust-bitcoin` project?

sgeisler:
> Do you think it's also time to raise our version to 1.29 to match the rust-bitcoin project?

Oh lol, I was totally unaware that we didn't do that yet, otoh no surprise given this crate's stability. If it helps sure! Idk how much better clippy can cope with 1.29 though, it's also ancient.

sgeisler (commented):
The new tests are a bit redundant (why `check` when you are going to build and test anyway?) and I'd prefer to only test if there's not a good argument to be made for checking. Later we will probably expand the matrix to test features like no-std and this could cost us time when we hit the maximum parallel worker limit.

src/lib.rs
clarkmoody: What does this do for us?

clarkmoody:
> The new tests are a bit redundant (why `check` when you are going to build and test anyway?) and I'd prefer to only test if there's not a good argument to be made for checking. Later we will probably expand the matrix to test features like no-std and this could cost us time when we hit the maximum parallel worker limit.

Agree. Removed the `check` job, as that functionality is duplicated by `test`.

sgeisler (approved):
ACK 335f6bbea8d6875fda3208c06e6c428f3e480bc3

src/lib.rs
sgeisler: It's necessary to implement the `Variant::Bech32 as u32` API afaik (making the enum castable to its internal representation). I don't really like it because it isn't too well-known this is possible and also seems a bit like abusing this feature here for probably non-existent performance gains.

clarkmoody: Makes sense. I'll leave this out for now, and we can circle back later if the need arises.

PR #58, opened 2021-11-25
Implement From<u5> for u8

tcharding:
Currently we have an implementation of `Into` for `u8`, if we implement
`From<u5>` for `u8` we get the `Into` implementation for free.

> One should avoid implementing Into and implement From instead. Implementing From automatically provides one with an implementation of Into thanks to the blanket implementation in the standard library.

ref: https://doc.rust-lang.org/std/convert/trait.Into.html

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -99,9 +99,9 @@ impl u5 {
     }
 }
 
-impl Into<u8> for u5 {
-    fn into(self) -> u8 {
-        self.0
+impl From<u5> for u8 {
+    fn from(v: u5) -> u8 {
+        v.0
     }
 }
 


Kixunil:
Actually, this breaks MSRV. I personally wouldn't mind if all Rust Bitcoin projects bumped MSRV to 1.41.1 but others want to keep it lower.

tcharding:
Why do you think it breaks MSRV? `Try` and `From` are available from Rust 1.0.0? Did you mix up this with `TryFrom` (which is 1.34.0)?

Kixunil:
`u8` is not from this crate and the very documentation you linked contains a whole [section explaining why it's a problem](https://doc.rust-lang.org/std/convert/trait.Into.html#implementing-into-for-conversions-to-external-types-in-old-versions-of-rust) :)

Kixunil:
Oh, wait, sorry, I think it *does* work for *non-generic* types.

Kixunil:
Yes, [exactly](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7ddc7c1c31c9f271c08e6144c5f3a0be), please disregard my previous objection

clarkmoody (approved):
LGTM!

PR #62, opened 2022-01-12
avoid potential double-panic

nappa85:
A drop can be caused by a panic, panicking inside a panic causes a double-panic, therefore an abort

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -71,7 +71,7 @@ use alloc::borrow::Cow;
 #[cfg(any(feature = "std", test))]
 use std::borrow::Cow;
 
-use core::{fmt, mem};
+use core::fmt;
 
 /// Integer in the range `0..32`
 #[derive(PartialEq, Eq, Debug, Copy, Clone, Default, PartialOrd, Ord, Hash)]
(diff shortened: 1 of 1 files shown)


clarkmoody:
So a panic can cause other `drop` methods to be called normally?

nappa85:
> So a panic can cause other `drop` methods to be called normally?

Stating the official docs https://doc.rust-lang.org/nightly/core/ops/trait.Drop.html#panics
```Given that a panic! will call drop as it unwinds, any panic! in a drop implementation will likely abort.```

tcharding:
Perhaps we should just not write the checksum from the `Drop` implementation, since there is an error that should be checked for, would it not be better for users to manually add the checksum so the error can be checked?

clarkmoody:
> Perhaps we should just not write the checksum from the `Drop` implementation, since there is an error that should be checked for, would it not be better for users to manually add the checksum so the error can be checked?

Yeah that sounds like a better approach. Maybe the write during Drop is a bit too much magic?

Kixunil:
Nah, writing on drop is fine, `BufWriter` from `std` does is too (ignoring errors) and we should probably get inspiration from there. Especially have [`panicked` field](https://doc.rust-lang.org/src/std/io/buffered/bufwriter.rs.html#80)

tcharding:
Needs rebase mate.

apoelstra:
Going to close this, since `Bech32Writer` is being dropped. But thank you!

PR #66, opened 2022-06-24
Support bech32 encoding without a checksum

jkczyz:
[BOLT 12 Offers](https://github.com/lightning/bolts/pull/798) uses bech32 encoding without a checksum since QR codes already have a checksum. Add functions `encode_without_checksum` and `decode_without_checksum` to support this use case.

Also, remove overall length check in decode since it is unnecessary.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -182,12 +182,12 @@ impl<'a> Bech32Writer<'a> {
 
     /// Write out the checksum at the end. If this method isn't called this will happen on drop.
     pub fn finalize(mut self) -> fmt::Result {
-        self.inner_finalize()?;
+        self.write_checksum()?;
         mem::forget(self);
         Ok(())
     }
 
-    fn inner_finalize(&mut self) -> fmt::Result {
+    fn write_checksum(&mut self) -> fmt::Result {
         // Pad with 6 zeros
         for _ in 0..CHECKSUM_LENGTH {
             self.polymod_step(u5(0))
@@ -203,6 +203,7 @@ impl<'a> Bech32Writer<'a> {
         Ok(())
     }
 }
+
 impl<'a> WriteBase32 for Bech32Writer<'a> {
     type Err = fmt::Error;
 
@@ -215,7 +216,7 @@ impl<'a> WriteBase32 for Bech32Writer<'a> {
 
 impl<'a> Drop for Bech32Writer<'a> {
     fn drop(&mut self) {
-        self.inner_finalize()
+        self.write_checksum()
             .expect("Unhandled error writing the checksum on drop.")
     }
 }
@@ -421,6 +422,32 @@ pub fn encode_to_fmt<T: AsRef<[u5]>>(
     })?)
 }
 
+/// Encode a bech32 payload without a checksum to an [fmt::Write].
+/// This method is intended for implementing traits from [std::fmt].
+///
+/// # Errors
+/// * If [check_hrp] returns an error for the given HRP.
+/// * If `fmt` fails on write
+/// # Deviations from standard
+/// * No length limits are enforced for the data part
+pub fn encode_without_checksum_to_fmt<T: AsRef<[u5]>>(
+    fmt: &mut fmt::Write,
+    hrp: &str,
+    data: T,
+) -> Result<(), Error> {
+    let hrp = match check_hrp(hrp)? {
+        Case::Upper => Cow::Owned(hrp.to_lowercase()),
+        Case::Lower | Case::None => Cow::Borrowed(hrp),
+    };
+
+    fmt.write_str(&hrp)?;
+    fmt.write_char(SEP)?;
+    for b in data.as_ref() {
+        fmt.write_char(b.to_char())?;
+    }
+    Ok(())
+}
+
 /// Used for encode/decode operations for the two variants of Bech32
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
 pub enum Variant {
@@ -463,15 +490,48 @@ pub fn encode<T: AsRef<[u5]>>(hrp: &str, data: T, variant: Variant) -> Result<St
     Ok(buf)
 }
 
+/// Encode a bech32 payload to string without the checksum.
+///
+/// # Errors
+/// * If [check_hrp] returns an error for the given HRP.
+/// # Deviations from standard
+/// * No length limits are enforced for the data part
+pub fn encode_without_checksum<T: AsRef<[u5]>>(hrp: &str, data: T) -> Result<String, Error> {
+    let mut buf = String::new();
+    encode_without_checksum_to_fmt(&mut buf, hrp, data)?;
+    Ok(buf)
+}
+
 /// Decode a bech32 string into the raw HRP and the data bytes.
 ///
-/// Returns the HRP in lowercase..
+/// Returns the HRP in lowercase, the data with the checksum removed, and the encoding.
 pub fn decode(s: &str) -> Result<(String, Vec<u5>, Variant), Error> {
-    // Ensure overall length is within bounds
-    if s.len() < CHECKSUM_LENGTH + 2 {
+    let (hrp_lower, mut data) = split_and_decode(s)?;
+    if data.len() < CHECKSUM_LENGTH {
         return Err(Error::InvalidLength);
     }
 
+    // Ensure checksum
+    match verify_checksum(hrp_lower.as_bytes(), &data) {
+        Some(variant) => {
+            // Remove checksum from data payload
+            data.truncate(data.len() - CHECKSUM_LENGTH);
+
+            Ok((hrp_lower, data, variant))
+        }
+        None => Err(Error::InvalidChecksum),
+    }
+}
+
+/// Decode a bech32 string into the raw HRP and the data bytes, assuming no checksum.
+///
+/// Returns the HRP in lowercase and the data.
+pub fn decode_without_checksum(s: &str) -> Result<(String, Vec<u5>), Error> {
+    split_and_decode(s)
+}
+
+/// Decode a bech32 string into the raw HRP and the `u5` data.
+fn split_and_decode(s: &str) -> Result<(String, Vec<u5>), Error> {
     // Split at separator and check for two pieces
     let (raw_hrp, raw_data) = match s.rfind(SEP) {
         None => return Err(Error::MissingSeparator),
@@ -480,9 +540,6 @@ pub fn decode(s: &str) -> Result<(String, Vec<u5>, Variant), Error> {
             (hrp, &data[1..])
         }
     };
-    if raw_data.len() < CHECKSUM_LENGTH {
-        return Err(Error::InvalidLength);
-    }
 
     let mut case = check_hrp(raw_hrp)?;
     let hrp_lower = match case {
@@ -492,7 +549,7 @@ pub fn decode(s: &str) -> Result<(String, Vec<u5>, Variant), Error> {
     };
 
     // Check data payload
-    let mut data = raw_data
+    let data = raw_data
         .chars()
         .map(|c| {
             // Only check if c is in the ASCII range, all invalid ASCII
@@ -527,17 +584,7 @@ pub fn decode(s: &str) -> Result<(String, Vec<u5>, Variant), Error> {
         })
         .collect::<Result<Vec<u5>, Error>>()?;
 
-    // Ensure checksum
-    match verify_checksum(hrp_lower.as_bytes(), &data) {
-        Some(variant) => {
-            // Remove checksum from data payload
-            let dbl: usize = data.len();
-            data.truncate(dbl - CHECKSUM_LENGTH);
-
-            Ok((hrp_lower, data, variant))
-        }
-        None => Err(Error::InvalidChecksum),
-    }
+    Ok((hrp_lower, data))
 }
 
 fn verify_checksum(hrp: &[u8], data: &[u5]) -> Option<Variant> {
@@ -801,6 +848,8 @@ mod tests {
                 Error::InvalidLength),
             ("1p2gdwpf",
                 Error::InvalidLength),
+            ("bc1p2",
+                Error::InvalidLength),
         );
         for p in pairs {
             let (s, expected_error) = p;
@@ -930,7 +979,7 @@ mod tests {
     }
 
     #[test]
-    fn writer() {
+    fn write_with_checksum() {
         let hrp = "lnbc";
         let data = "Hello World!".as_bytes().to_base32();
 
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -134,6 +134,7 @@ pub struct Bech32Writer<'a> {
     formatter: &'a mut fmt::Write,
     chk: u32,
     variant: Variant,
+    checksum: Checksum,
```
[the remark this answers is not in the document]
jkczyz: Made a similar wrapper for without a checksum, though there's a small amount of code duplication that way.

tcharding:
Sure thing. Can you squash the first two commits please because the second one is patching the changes in the first. Not super important but I'd put patch 3 at the front since it is a trivial improvement. Thanks.

tcharding:
Hi @jkczyz, thanks for your contribution. This PR does the job but its not pretty. How urgent is this? Is it something you need to use today or is it just something you did for fun?

The reason its not pretty is that it adds functionality that is obviously just plastered on top. By that I mean the whole `Bech32Writer` design is there to track and add the checksum so then adding another type that mirrors it but does not add a checksum is kinda strange. I'm basically making the same observation @TheBlueMatt did above but on this iteration of the PR. If we want a good solution I think we need to dig a bit deeper.

jkczyz:
> Hi @jkczyz, thanks for your contribution. This PR does the job but its not pretty. How urgent is this? Is it something you need to use today or is it just something you did for fun?
> 

Somewhat urgent as it will hold up our work on Offers (https://github.com/lightningdevkit/rust-lightning/pull/1597). I have a draft that uses this through `rust-bitcoin`'s re-export, so would need a new release for both.

> The reason its not pretty is that it adds functionality that is obviously just plastered on top. By that I mean the whole `Bech32Writer` design is there to track and add the checksum so then adding another type that mirrors it but does not add a checksum is kinda strange. I'm basically making the same observation @TheBlueMatt did above but on this iteration of the PR. If we want a good solution I think we need to dig a bit deeper.

I don't actually need to use `Bech32WithoutChecksumWriter` directly, so I'm happy to inline the code. I mainly included for  consistency and for code reuse, which is no longer the case it seems.

apoelstra:
As a stop-gap you could just drop the last 6 characters/u8s from a normal bech32 encoding.

jkczyz:
> Sure thing. Can you squash the first two commits please because the second one is patching the changes in the first. Not super important but I'd put patch 3 at the front since it is a trivial improvement. Thanks.

All done.

jkczyz:
> As a stop-gap you could just drop the last 6 characters/u8s from a normal bech32 encoding.

For encoding that would work, but we would still fail to decode a bech32 offer since it doesn't have a checksum.

src/lib.rs
```diff
@@ -471,15 +437,20 @@ pub fn encode_without_checksum_to_fmt<T: AsRef<[u5]>>(
...
-    }
+    let mut write = || {
+        fmt.write_str(&hrp)?;
+        fmt.write_char(SEP)?;
+        for b in data.as_ref() {
+            fmt.write_char(b.to_char())?;
+        }
+        Ok(())
+    };
+    Ok(write())
```
jkczyz, line 453: I'm a little uncertain about this. Seems to maintain the same behavior, but it is odd that an `fmt::Error` is always wrapped with `Ok`.

apoelstra: Yeah, the existing API is really weird.

apoelstra: Why are you constructing a closure and then immediately calling it?

jkczyz: Mainly as a simple way to maintain the API behavior. i.e., wrapping a `Err(fmt::Error)` returned in each of the three places using `?` with an `Ok`.  Without the closure, it would be:

```rust
    if let Err(e) = fmt.write_str(&hrp) { return Ok(Err(e)); }
    if let Err(e) = fmt.write_char(SEP) { return Ok(Err(e)); }
    for b in data.as_ref() {
        if let Err(e) = fmt.write_char(b.to_char()) { return Ok(Err(e)); }
    }
    Ok(Ok(()))
```

Happy to update the error type is your prefer. It probably should be an enum that is either `bech32::Error` or `fmt::Error`. Or maybe adding a variant to `bech32::Error` for `fmt::Error`.

apoelstra: This whole `Result<Result>` thing is weird to me -- I think we should add `fmt::Error` to the main `Error` enum and then flatten this return type.

I wouldn't worry about maintaining the API here. We want to overhaul this API anyway.

jkczyz: Flatted the `Result`. Let me know if you prefer a different name for the variant or if you want to drop the wrapped `fmt::Error` as it currently doesn't provide any additional information.

apoelstra:
Loosk good to me! Though I guess CI wants you to add a rustfmt commit.

@Kixunil mind taking another look at this? I think now my remaining unease with the API is just my existing unease at the API, and if we want to clean it up we should just go over the whole library.

ACK 3e95975d7c90450c2735a8cac1c6e95277713371

apoelstra (approved):
ACK 573d7f4756ed99f996e8ed9b3a70477fa12a3443

Kixunil:
Can't right now, give me a day or so.

src/lib.rs
```diff
@@ -460,19 +486,48 @@ impl Variant {
...
 }
 
 /// Decode a bech32 string into the raw HRP and the data bytes.
 ///
-/// Returns the HRP in lowercase..
+/// Returns the HRP in lowercase, the data with the checksum removed, and the encoding.
 pub fn decode(s: &str) -> Result<(String, Vec<u5>, Variant), Error> {
-    // Ensure overall length is within bounds
-    if s.len() < 8 {
+    let (hrp_lower, mut data) = decode_without_checksum(s)?;
```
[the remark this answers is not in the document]
tcharding: I convinced myself everything was ok by adding a test (that passes):
```rust

    #[test]
    fn roundtrip_without_checksum() {
        let hrp = "lnbc";
        let data = "Hello World!".as_bytes().to_base32();

        let encoded = encode_without_checksum(hrp, data.clone()).expect("failed to encode");
        let (decoded_hrp, decoded_data) = decode_without_checksum(&encoded).expect("failed to decode");

        assert_eq!(decoded_hrp, hrp);
        assert_eq!(decoded_data, data);
    }
```

Perhaps the `decode_without_checksum` public function should just call straight through to a helper function `_decode` so the `decode` function can use `_decode` as well and then it will read nicely? Something like

```rust

/// Decode a bech32 string into the raw HRP and the data bytes, assuming no checksum.
///
/// Returns the HRP in lowercase and the data with the checksum removed.
fn decode_without_checksum(s: &str) -> Result<(String, Vec<u5>), Error> {
    _decode(s)
}

/// Decode a bech32 string into the raw HRP and the `u5` data.
pub fn _decode(s: &str) -> Result<(String, Vec<u5>), Error> {
...
```

src/lib.rs
```diff
@@ -460,19 +486,48 @@ impl Variant {
...
+
+            Ok((hrp_lower, data, variant))
+        }
+        None => Err(Error::InvalidChecksum),
+    }
+}
+
+/// Decode a bech32 string into the raw HRP and the data bytes, assuming no checksum.
+///
+/// Returns the HRP in lowercase and the data with the checksum removed.
```
tcharding, line 529: This line is stale.

tcharding (approved):
Thanks for sticking with this! Functionality looks correct to me. Perhaps add the unit test I used if you like it. I'm not super happy with the function naming but perhaps we can leave tat till another day, like others have said already this crate is about to get an overhaul anyways. If this solves your usecase I'm happy to ack it.

tcharding:
tACK 573d7f4756ed99f996e8ed9b3a70477fa12a3443

jkczyz:
Addressed all feedback. Can squash fixups whenever you're ready.

apoelstra (approved):
ACK f768f039b0e136b9a90f711301d3dba0f997c770

apoelstra:
Looks good to me. Yes, would prefer squashing all the fixups.

apoelstra (approved):
ACK 86edca9371a63dfbce9fed27c00b214b936c19e3

src/lib.rs
```diff
@@ -213,7 +216,7 @@ impl<'a> WriteBase32 for Bech32Writer<'a> {
 
 impl<'a> Drop for Bech32Writer<'a> {
     fn drop(&mut self) {
-        self.inner_finalize()
+        self.write_checksum()
             .expect("Unhandled error writing the checksum on drop.")
```
Kixunil, line 220: FTR `std` ignores errors instead of panicking in drop and it is considered the right thing to do because panicking in drop is a footgun.

src/lib.rs
```diff
@@ -398,28 +401,51 @@ fn check_hrp(hrp: &str) -> Result<Case, Error> {
...
+/// * No length limits are enforced for the data part
+pub fn encode_without_checksum_to_fmt<T: AsRef<[u5]>>(
+    fmt: &mut fmt::Write,
+    hrp: &str,
+    data: T,
+) -> Result<(), Error> {
+    let hrp = match check_hrp(hrp)? {
+        Case::Upper => Cow::Owned(hrp.to_lowercase()),
+        Case::Lower | Case::None => Cow::Borrowed(hrp),
+    };
```
Kixunil, line 441: It is possible to o this without `Cow` which should be more efficient:

```rust
let hrp_owned; // yes, this is allowed in Rust!
let hrp = match check_hrp(hrp)? {
    Case::Upper => {
        hrp_owned = hrp.to_lowercase();
        hrp_owned.as_str()
    },
    Case::Lower | Case::None => hrp,
};
```

src/lib.rs
```diff
@@ -398,28 +401,51 @@ fn check_hrp(hrp: &str) -> Result<Case, Error> {
...
+/// This method is intended for implementing traits from [std::fmt].
+///
+/// # Errors
+/// * If [check_hrp] returns an error for the given HRP.
+/// * If `fmt` fails on write
+/// # Deviations from standard
+/// * No length limits are enforced for the data part
+pub fn encode_without_checksum_to_fmt<T: AsRef<[u5]>>(
+    fmt: &mut fmt::Write,
+    hrp: &str,
```
Kixunil, line 435: It'd be better to have `Hrp` newtype to make the reurn type `fmt::Result` allowing use in `Display` implementations.

src/lib.rs
```diff
@@ -622,6 +668,14 @@ pub enum Error {
     InvalidPadding,
     /// The whole string must be of one case
     MixedCase,
+    /// Writing UTF-8 data failed
+    WriteFailure(fmt::Error),
```
Kixunil, line 672: I don't think this is the right way of using `fmt::Error`. It's just a marker that something failed and you're supposed to retrieve the error fro underlying stream. In case of `std::io` it's done for you and in case of `String` there's no error so unwrapping is the correct thing to do.

This library abuses the `fmt::Write` API to signal display failure but `Display` must not fail for any other reason than formatter failing.

From [the docs](https://doc.rust-lang.org/std/fmt/index.html):

> Formatting implementations should ensure that they propagate errors from the [Formatter](https://doc.rust-lang.org/std/fmt/struct.Formatter.html) (e.g., when calling [write!](https://doc.rust-lang.org/std/macro.write.html)). However, they should never return errors spuriously. That is, a formatting implementation must and may only return an error if the passed-in [Formatter](https://doc.rust-lang.org/std/fmt/struct.Formatter.html) returns an error. This is because, contrary to what the function signature might suggest, string formatting is an infallible operation. This function only returns a result because writing to the underlying stream might fail and it must provide a way to propagate the fact that an error has occurred back up the stack.

Kixunil (commented):
Finally had the time to review it. The API has problems but it was the case before, so not a problem with this particular PR. It just needs to be cleaned up before 1.0.

src/lib.rs
apoelstra: Lol, oops! I tried to look for double-panics but got a bit confused and somehow missed this completely obvious one.

But I think we should remove this `Drop` business entirely so it's fine for now. I've never heard of a wrapper struct that takes a `&mut` ref and then has finalization logic on `Drop`...I suspect this was an API experiment that never failed enough to be removed.

src/lib.rs
apoelstra: That's a neat lifetime-organizing trick!

src/lib.rs
apoelstra: Good find! This will definitely inform our API overhaul.

src/lib.rs
Kixunil: IDK, it's not too different from `BufWriter` behavior. But I guess compiler warning would be better than drop. I was thinking of abusing `must_use` but then we couldn't implement any methods taking `&mut self` and it'd be fragile anyway.

jkczyz:
> > As a stop-gap you could just drop the last 6 characters/u8s from a normal bech32 encoding.
> 
> For encoding that would work, but we would still fail to decode a bech32 offer since it doesn't have a checksum.

For sake of planning, what timeframe should I expect for seeing this in a release? Do let me know if I can help in review if other PRs need to be merged first. Just want to set appropriate expectations given this will need to get into a `rust-bitcoin` release, too, before we can use it in `rust-lightning`.

tcharding:
We just dropped a `rust-bitcoin` release so the next one could be six months away. This poses an interesting dilemma, we don't want to be doing major releases of `rust-bitcoin` ever week but if a user is waiting on a merged feature in a dependency that we control it seems rude to hold it up too long.  

If we are going to release then #68 and #72 would likely be nice to consider first.

apoelstra:
We can do a much faster release of this than six months. It's still fine if rust-bitcoin depends on an old version -- `bech32` is pretty self-contained so it should be easy to work with multiple versions of it at once.

apoelstra:
But having said that yes, I'd like to do a bunch more work on this crate before releasing .. so it could be a month or more.

jkczyz:
Gotcha. I think had trouble with multiple versions but can't recall the exact error. Will try again once there's a new `bech32` release. Otherwise, I can unblock the current work by holding off on offer parsing. I made a builder framework in the interim, so we can move forward with dependent work using that.

Kixunil:
Could the code be changed to be non-breaking? Minor version bump wouldn't be so painful and we could do API overhaul later.

jkczyz:
> Could the code be changed to be non-breaking? Minor version bump wouldn't be so painful and we could do API overhaul later.

Are you suggesting doing that in `rust-bech32` and do a similar minor version bump in `rust-bitcoin`?

jkczyz:
Ah, so since `rust-bitcoin` specifies `version = "0.9.0"` for `bech32`, any version `>=0.9.0, <0.10.0` cab be used. Sure, I can put together a PR.

jkczyz:
Opened #77.

PR #74, opened 2022-07-29
Split out a `u5` crate

tcharding:
Split out a `u5` crate. This is draft because it includes a bunch of open PRs. It is mainly to showcase the improvements I did to `u5` based on the Rust stdlib for `u8`.

I don't know if this will ever be useful but if we are going to make `u5` 1.0 then in it seems reasonable to have all this stuff.


Ericson2314:
I like this idea! I hope someday Rust will get built-in non-round-number-byte numbers like Erlang; it is great to do this in the meantime to demonstrate this is useful.

tcharding:
I've gone over this again and I believe it now includes _most_ functionality of Rust stdlib unsigned types, I skipped some of the no-op methods (e.g., `overflowing_rem`). If the other open PRs merge I'll rebase this and take it off draft. This PR will be pretty time consuming to review though.

PR #76, opened 2022-08-11
Install gcc-arm-none-eabi in embedded ci job

tcharding:
The current embedded ci job is failing, I think its because we need to
install `gcc-arm-none-eabi`, based on how we do it in `rust-bitcoin`.


tcharding:
BOOM! This will fix all the red in other PR ci runs.

apoelstra (approved):
ACK 091240e2f9cd1f440c202f48013e1843e5978feb

PR #77, opened 2022-08-12
Revert `encode_to_fmt` API change

jkczyz:
Maintaining the original error type allows for making a minor release since it won't break the API. This lets indirect consumers of the crate through `rust-bitcoin`'s module re-export to use new functions (e.g., `encode_without_checksum`) without a new `rust-bitcoin` release.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bech32"
-version = "0.9.0"
+version = "0.9.1"
 authors = ["Clark Moody"]
 repository = "https://github.com/rust-bitcoin/rust-bech32"
 description = "Encodes and decodes the Bech32 format"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -401,7 +401,6 @@ fn check_hrp(hrp: &str) -> Result<Case, Error> {
 ///
 /// # Errors
 /// * If [check_hrp] returns an error for the given HRP.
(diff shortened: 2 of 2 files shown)


jkczyz:
Should this be merged into `master` and later reverted? Or into a separate release branch?

apoelstra:
I think we should just merge into master and later revert it. No need for extra process for such a low-activity crate.

apoelstra:
Can you rebase now that #76 is merged?

apoelstra (approved):
ACK 36939353de20abdd65744e87ca4e365b0f4278cc

apoelstra:
If you add a commit which bumps the minor version number I can publish a release right after merging.

apoelstra (approved):
ACK 0d01fadfc4e32853ab6aff18a51b062d33630179

apoelstra:
Thanks! Tagged, signed and published.

jkczyz:
> Thanks! Tagged, signed and published.

Thanks for the quick turnaround on this! I had to add an explicit dependency in our crate otherwise I'd still get 0.9.0 from `rust-bitcoin` even after a `cargo clean`. But no conflicting definitions this time.

apoelstra:
Did you try `cargo update`?

jkczyz:
Ah, right. 😅 Forgot the `Cargo.lock` file dictates which version is used. Works without the explicit dependency.

Issue #80, opened 2022-10-01
`u5` does not implement `Display`


tcharding:
Any reason why the `u5` type does not implement `Display`?

tcharding:
Closing as this was implicitly resolved by the recent crate re-write.

PR #89, opened 2023-03-03
implement Display for u5

tcharding:
No reason not to enable printing u5s.

Fix: #80


apoelstra:
It looks like these are being displayed as decimal numbers. This is unexpected. They should be displayed as bech32 characters.

tcharding:
bech32 knows about `u5` but `u5` does not know about bech32; isn't `u5` a stand alone type and hence its `Display` impl should behave like other standard integer types?

tcharding:
Review question unresolved but I rebased while thinking about it.

apoelstra:
I think of `u5` as a field element in GF32, definitely *not* as a standalone integer type. We could rename it if this is confusing.

Do we have `Add` and `Mul` implemented on this type, and do they respect field operations or "mod 32" operatons?

tcharding:
Closing this, see https://github.com/rust-bitcoin/rust-bech32/issues/94

Issue #82, opened 2023-02-03
`impl<T: AsRef<[u8]>> ToBase32 for T` should be `impl<T: AsRef<[u8]> + ?Sized> ToBase32 for T`


Kixunil:
The current implementation leads to confusing type errors and needlessly limits the usable types.

PR #90, opened 2023-03-03
Do not require know size at compile time

tcharding:
The `Base32Len` and `ToBase32` traits currently require the implementing type to have a known size at compile time. This is an unnecessary restriction.

Fix: #82

FTR I did not know at first which functions this reasoning could by applied to so I added `+ ?Sized` to every generic that was `AsRef<[u8]>` and let the compiler tell me which ones were correct.


apoelstra (approved):
ACK 060ef26b3e3911ce0434864a87dbc73ca3482299

Issue #60, opened 2022-01-05
Polishing/stabilization roadmap?


Kixunil:
It'd be great to polish this crate and make it stable with idiomatic API and good documentation. Since this crate appears stale I'm willing to help maintain it to make the progress a bit faster.

I'm still unsure what all those traits do, they seem to be meant as extension traits?
Thinking about ideal API, I identified these basic building blocks:

* `u5` type
* converting stream of `u8` to stream of `u5` and back
* calculating checksum
* converting stream of `u8` to ASCII and back

I think each deserves its own module, maybe even crate. The commonly used things can still be flattened for consumers.

### u5 type

I think this is good as is. Could add `impl TryFrom<*> for u5` and arithmetic functions for completeness.

### u8-u5 conversion

The API should enable to perform this conversion without allocation by having wrappers around iterators.
Convenience functions for slices and `Vec`s can be added but not sure how important given `.iter()` and `.collect()` are easy to call.
This should also provide free-standing functions to calculate encoded/decoded lengths.

### checksum calculation

The API should be similar to hashers with possible convenience to compute from iterators.
It should also provide a wrapper for the case when checksum is calculated together with conversions.

### u8-ASCII conversion

Should be similar to `u8`-`u5` conversion. Might be useful to provide wrappers returning both ASCII `u8` and `char` to avoid overhead when writing to `writer`. Also maybe provide a method to construct `String` without unnecessary UTF-8 check (using `unsafe` internally).

### Facade

Convenience functions for converting stream of `u8` to strings (with/without checksum) and back should be provided.

Also for consideration: exploit the invariant of `u5` to skip bounds checks when converting to `char`.

tcharding:
Would be great to get your input @Kixunil on how close you think the new code is to being ready for v1.0.0.

Current plan is (only from informal discussion between @apoelstra and myself, so still requires input from @clarkmoody):

1. Merge #128
2. Release v0.10.0
3. Let that sit in the wild for six months or so

Seems we are then pretty close to doing a v1.0.0

Kixunil:
Great timing, I'm back!

I only took a quick look now and I still have a bunch of things.

* Why do we have `WriteBase256`? Looks like it should be just the `Write` trait, possibly from `lgio`
* `_unchecked` in Rust usually means "UB if wrong" or "panic if wrong" but has different semantics here. Rename it?
* Didn't look into error types yet but I think we should have some global standard around them.
[...]

PR #92, opened 2023-03-03
Implement `TryFrom<T> for u5` for all standard integer types

tcharding:
Patch 1 is preparatory clean up.

From the commit log of patch 2:

    Currently we use `Error::InvalidData(u8)` to attempt to cache an invalid
    value encountered while converting to a `u8`. This is buggy because we
    cast to the `u8` to stash the value so we loose valuable bits.
    
    Implement `TryFrom` by doing:
    - Add an `Error::Overflow` variant.
    - Use `Error::Overflow` for conversion _outside_ of `TryFrom` impls
    - Add a new `TryFromError` and nest it in `Error::TryFrom`
    - Implement `TryFrom<T> for u5` impls for all standard integer types.
    
    Inspired by issue #60 and the stdlib macros of the same name.

Please note, there is expected error work to come soon on this crate, can we keep perfect error code out of scope for this PR please (obviously if it has gaping problems then yell at me)? Note that the same derives are used here as for `Error` even though I think they are wrong.

Done as part of stabilising (#60).

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,0 +1,31 @@
+// Written by the Rust Bitcoin developers.
+// SPDX-License-Identifier: CC0-1.0
+
+// TODO: We can get this from `bitcoin-internals` crate once that is released.
+
+//! # Error
+//!
+//! Error handling macros and helpers.
+//!
+
+/// Formats error.
+///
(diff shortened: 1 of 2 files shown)


apoelstra (approved):
ACK 25462c9bbdc6684764b4d5dc3f259ecc4d2baf1f

PR #93, opened 2023-03-21
Run fuzzer in CI

tcharding:
As we do in `rust-bitcoin` run the fuzzer for each PR and each push.

Fix: #43


apoelstra (approved):
ACK 37d285ce01cca099ad96e1c11d02ac04fd4ed61b

PR #98, opened 2023-04-06
Add gf32 module

tcharding:
We are attempting to get this crate to v0.1 (ie, API stabilization).

Currently we use a type call `u5` for the 5 bit math, while valid this is slightly misleading because our `u5` is not basic integer type. Within the bech32 encoding scheme the 5 bit type is an element of GF32 (Galois Field).

Add a `gf32` module that defines a type `Fe32` (Field Element). This type is similar to the current `u5` but includes add/sub/div/mul over the finite field.

Introduce the `primitives` module to hold the `gf32` module (and other future primitive types).

As yet do not make use of the `Fe32` type. Doing so is trivial and will be done later after more primitive types are added.

### Note on attribution

I'd like to be conscious not to make you feel like we are clobbering your crate clarkmoody, as such I left an attribution in this new file to you even though I copied the code from another crate that apoelstra had written. I felt this was correct because if this merges and eventually replaces the `u5` type the code is really similar to what you wrote originally. Also we are keeping your chosen MIT license and re licensing any code we pull from other places (I'm speaking for apoelstra really because the CC license is his favoured license but I believe I understand his wishes on this issue since we have discussed it before).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,13 +1,13 @@
 [package]
 name = "bech32"
 version = "0.9.1"
-authors = ["Clark Moody", "The rust-bitcoin developers"]
+authors = ["Clark Moody", "Andrew Poelstra", "Tobin Harding", "The rust-bitcoin developers"]
 repository = "https://github.com/rust-bitcoin/rust-bech32"
 documentation = "https://docs.rs/bech32/"
-description = "Encodes and decodes the Bech32 format"
+description = "Encodes and decodes the Bech32 format and implements the bech32 and bech32m checksums"
 readme = "README.md"
-keywords = ["base32", "encoding", "bech32"]
-categories = ["encoding"]
+keywords = ["base32", "encoding", "bech32", "bech32m", "bitcoin", "cryptocurrency"]
+categories = ["encoding", "cryptography::cryptocurrencies"]
 license = "MIT"
 edition = "2018"
 
@@ -20,3 +20,6 @@ alloc = []
 version = "0.7.1"
 default-features = false
 optional = true
+
+[target.'cfg(mutate)'.dev-dependencies]
+mutagen = { git = "https://github.com/llogiq/mutagen" }
--- a/src/primitives/gf32.rs
+++ b/src/primitives/gf32.rs
@@ -1,0 +1,470 @@
+// Written by the Andrew Poelstra and the rust-bitcoin developers.
+// (Replacing very similar code originally written by Clark Moody.)
+// SPDX-License-Identifier: MIT
+
+//! GF32 - Galois Field over 32 elements.
+//!
+//! Implements GF32 arithmetic, defined and encoded as in BIP-0173 "bech32".
+//!
+//! > A finite field is a finite set which is a field; this means that multiplication, addition,
+//! > subtraction and division (excluding division by zero) are defined and satisfy the rules of
+//! > arithmetic known as the field axioms.
+//!
+//! ref: <https://en.wikipedia.org/wiki/Finite_field>
+
+use core::convert::{Infallible, TryFrom};
+use core::{fmt, num, ops};
+
+#[cfg(all(test, mutate))]
+use mutagen::mutate;
+
+/// Logarithm table of each bech32 element, as a power of alpha = Z.
+///
+/// Includes Q as 0 but this is false; you need to exclude Q because it has no discrete log. If we
+/// could have a 1-indexed array that would panic on a 0 index that would be better.
+#[rustfmt::skip]
+const LOG: [isize; 32] = [
+     0,  0,  1, 14,  2, 28, 15, 22,
+     3,  5, 29, 26, 16,  7, 23, 11,
+     4, 25,  6, 10, 30, 13, 27, 21,
+    17, 18,  8, 19, 24,  9, 12, 20,
+];
+
+/// Mapping of powers of 2 to the numeric value of the element.
+#[rustfmt::skip]
+const LOG_INV: [u8; 31] = [
+     1,  2,  4,  8, 16,  9, 18, 13,
+    26, 29, 19, 15, 30, 21,  3,  6,
+    12, 24, 25, 27, 31, 23,  7, 14,
+    28, 17, 11, 22,  5, 10, 20,
+];
+
+/// Mapping from numeric value to bech32 character.
+#[rustfmt::skip]
+const CHARS_LOWER: [char; 32] = [
+    'q', 'p', 'z', 'r', 'y', '9', 'x', '8', //  +0
+    'g', 'f', '2', 't', 'v', 'd', 'w', '0', //  +8
+    's', '3', 'j', 'n', '5', '4', 'k', 'h', // +16
+    'c', 'e', '6', 'm', 'u', 'a', '7', 'l', // +24
+];
+
+/// Mapping from bech32 character (either case) to numeric value.
+///
+/// E.g., 'z' is CHARS_LOWER[2] and is ASCII value 122 so CHARS_INV[122] == 2
+#[rustfmt::skip]
+const CHARS_INV: [i8; 128] = [
+    -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+    -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+    -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+    15, -1, 10, 17, 21, 20, 26, 30,  7,  5, -1, -1, -1, -1, -1, -1,
+    -1, 29, -1, 24, 13, 25,  9,  8, 23, -1, 18, 22, 31, 27, 19, -1,
+     1,  0,  3, 16, 11, 28, 12, 14,  6,  4,  2, -1, -1, -1, -1, -1,
+    -1, 29, -1, 24, 13, 25,  9,  8, 23, -1, 18, 22, 31, 27, 19, -1,
+     1,  0,  3, 16, 11, 28, 12, 14,  6,  4,  2, -1, -1, -1, -1, -1,
+];
+
+/// An element in GF(32), the finite field containing elements `[0,31]` inclusive.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+pub struct Fe32(pub(crate) u8);
+
+impl Fe32 {
+    // These are a little gratuitous for a reference implementation, but it makes me happy to do it.
+    /// Numeric value maps to bech32 character: 0 == "q".
+    pub const Q: Fe32 = Fe32(0);
+    /// Numeric value maps to bech32 character: 1 == "p".
+    pub const P: Fe32 = Fe32(1);
+    /// Numeric value maps to bech32 character: 2 == "z".
+    pub const Z: Fe32 = Fe32(2);
+    /// Numeric value maps to bech32 character: 3 == "r".
+    pub const R: Fe32 = Fe32(3);
+    /// Numeric value maps to bech32 character: 4 == "y".
+    pub const Y: Fe32 = Fe32(4);
+    /// Numeric value maps to bech32 character: 5 == "9".
+    pub const _9: Fe32 = Fe32(5);
+    /// Numeric value maps to bech32 character: 6 == "x".
+    pub const X: Fe32 = Fe32(6);
+    /// Numeric value maps to bech32 character: 7 == "8".
+    pub const _8: Fe32 = Fe32(7);
+    /// Numeric value maps to bech32 character: 8 == "g".
+    pub const G: Fe32 = Fe32(8);
+    /// Numeric value maps to bech32 character: 9 == "f".
+    pub const F: Fe32 = Fe32(9);
+    /// Numeric value maps to bech32 character: 10 == "2".
+    pub const _2: Fe32 = Fe32(10);
+    /// Numeric value maps to bech32 character: 11 == "t".
+    pub const T: Fe32 = Fe32(11);
+    /// Numeric value maps to bech32 character: 12 == "v".
+    pub const V: Fe32 = Fe32(12);
+    /// Numeric value maps to bech32 character: 13 == "d".
+    pub const D: Fe32 = Fe32(13);
+    /// Numeric value maps to bech32 character: 14 == "w".
+    pub const W: Fe32 = Fe32(14);
+    /// Numeric value maps to bech32 character: 15 == "0".
+    pub const _0: Fe32 = Fe32(15);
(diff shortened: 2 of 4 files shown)


tcharding:
Andrew and I have gotten pretty psyched over the last few days and hacked up a bunch of stuff, if anyone is interested to look, check out `04-04-primitives-api` branch on my `rust-bech32` tree. Its WIP, but that is the direction we are hoping to push.

apoelstra:
cc @clarkmoody we won't merge this without your signoff, since it's part of a larger project that substantially rewrites the library in terms of iterator adaptors.

(I edited the PR description to remove all the @ signs.)

Cargo.toml
```diff
@@ -1,12 +1,12 @@
 [package]
 name = "bech32"
 version = "0.9.1"
-authors = ["Clark Moody", "The rust-bitcoin developers"]
+authors = ["Clark Moody", "Andrew Poelstra", "Tobin Harding", "The rust-bitcoin developers"]
 repository = "https://github.com/rust-bitcoin/rust-bech32"
 documentation = "https://docs.rs/bech32/"
-description = "Encodes and decodes the Bech32 format"
+description = "Encodes and decodes the Bech32 format (bech32 and bech32m)"
```
apoelstra, line 7: In 310012dd3e5f5725b5bde9021606ed1d5542620d:

I'd say "and implement the bech32 and bech32m checksums" rather than putting these in parens.

Cargo.toml
```diff
@@ -1,12 +1,12 @@
...
 readme = "README.md"
-keywords = ["base32", "encoding", "bech32"]
+keywords = ["base32", "encoding", "bech32", "bech32m"]
```
apoelstra, line 9: Maybe also add "bitcoin" and "cryptocurrency" here. I forget if there's a limit on the number of keywords you can have.

src/primitives/gf32.rs
```diff
@@ -0,0 +1,481 @@
...
+    pub const _0: Fe32 = Fe32(15);
+    /// Numeric value maps to bech32 character: 16 == "s".
+    pub const S: Fe32 = Fe32(16);
+    /// Numeric value maps to bech32 character: 17 == "3".
+    pub const _3: Fe32 = Fe32(17);
+    /// Numeric value maps to bech32 character: 18 == "j".
+    pub const J: Fe32 = Fe32(18);
+    /// Numeric value maps to bech32 character: 19 == "n".
+    pub const N: Fe32 = Fe32(19);
+    /// Numeric value maps to bec32 character: 20 == "5".
```
apoelstra, line 110: In 310012dd3e5f5725b5bde9021606ed1d5542620d:

lol multiple of these misspell `bech32`

src/primitives/gf32.rs
```diff
@@ -0,0 +1,481 @@
...
+    15, -1, 10, 17, 21, 20, 26, 30,  7,  5, -1, -1, -1, -1, -1, -1,
+    -1, 29, -1, 24, 13, 25,  9,  8, 23, -1, 18, 22, 31, 27, 19, -1,
+     1,  0,  3, 16, 11, 28, 12, 14,  6,  4,  2, -1, -1, -1, -1, -1,
+    -1, 29, -1, 24, 13, 25,  9,  8, 23, -1, 18, 22, 31, 27, 19, -1,
+     1,  0,  3, 16, 11, 28, 12, 14,  6,  4,  2, -1, -1, -1, -1, -1,
+];
+
+/// An element of GF32.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+pub struct Fe32(pub(crate) u8); // pub(crate) because of `hrpstring` module.
```
apoelstra, line 66: In 310012dd3e5f5725b5bde9021606ed1d5542620d:

I'd remove this comment. Every module in `primitives` will probably use the internal `u8`.

src/primitives/gf32.rs
```diff
@@ -0,0 +1,481 @@
...
+        let byte = i8::try_from(u32::from(c)).map_err(|_| Error::InvalidChar(c))?;
+        let byte = byte as u8; // cast guaranteed to be ok since we started with an unsigned value
+        let u5 = u8::try_from(CHARS_INV[usize::from(byte)]).map_err(|_| Error::InvalidChar(c))?;
+        Ok(Fe32(u5))
+    }
+
+    /// Converts the field element to a lowercase bech32 character
+    pub fn to_char(self) -> char {
+        // casting and indexing fine as we have self.0 in [0, 32) as an invariant
+        CHARS_LOWER[self.0 as usize]
```
apoelstra, line 186: In 310012dd3e5f5725b5bde9021606ed1d5542620d:

I think we can remove the cast and do `.into`. Since Rust 1.5 we've been able to convert from u8 this way (and since 1.26 from u16).

src/primitives/gf32.rs
```diff
@@ -0,0 +1,481 @@
...
+    }
+
+    /// Converts the field element to a 5-bit u8, with bits representing the coefficients
+    /// of the polynomial representation.
+    pub fn to_u8(self) -> u8 { self.0 }
+
+    fn _add(self, other: Fe32) -> Fe32 { Fe32(self.0 ^ other.0) }
```
apoelstra, line 193: In 310012dd3e5f5725b5bde9021606ed1d5542620d:

Why do we have these methods when we already have operator overloads?

src/primitives/gf32.rs
```diff
@@ -0,0 +1,481 @@
...
+            Fe32(LOG_INV[((31 + log1 - log2) % 31) as usize])
+        }
+    }
+}
+
+impl fmt::Display for Fe32 {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { fmt::Display::fmt(&self.to_char(), f) }
+}
+
+impl str::FromStr for Fe32 {
```
apoelstra, line 227: In 310012dd3e5f5725b5bde9021606ed1d5542620d:

IMO we should just drop this impl. Converting from a *string* to a single field element is a super weird thing to do, and I don't use this anywhere in my code (because I'd forgotten I"d implemented it :)).

Dropping this would also let us drop the `ExtraChar` and `InvalidLength` variants from the Error enum.

src/primitives/gf32.rs
```diff
@@ -0,0 +1,481 @@
...
+                /// Tries to create a [`Fe32`] type from a signed source number type.
+                ///
+                /// # Errors
+                ///
+                /// Returns an error if `value` is outside of the range of an `Fe32`.
+                fn try_from(value: $ty) -> Result<Self, Self::Error> {
+                    if value.is_negative() {
+                        Err(Error::Negative)?;
+                    }
+                    let byte = u8::try_from(value)?;
```
apoelstra, line 280: I think the negative check is entirely superfluous since `u8::try_from` will do it for us. So we can drop that, and drop the `Error::Negative` variant.

src/primitives/gf32.rs
```diff
@@ -0,0 +1,481 @@
...
+        // 1. Produce the translation wheel by multiplying
+        let logbase = Fe32(20);
+        let mut init = Fe32(1);
+        let mut s = String::new();
+        for _ in 0..31 {
+            s.push(init.to_char());
+            init *= logbase;
+        }
+        // Can be verified against the multiplication disk, starting with P and moving
+        // clcockwise
```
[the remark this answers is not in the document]
apoelstra: Also we should maybe link to the Codex32 booklet https://github.com/BlockstreamResearch/codex32/blob/master/SSS32.ps when citing "the multiplication disk".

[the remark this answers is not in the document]
apoelstra: I guess on a comment at the top of this entire unit test. And on `recovery_wheel`.

apoelstra:
Done reviewing 310012dd3e5f5725b5bde9021606ed1d5542620d. Sorry for having so many nits on code that I largely wrote myself :P but I didn't really look at it after writing it.

Cargo.toml
tcharding: Added. I added this too:
`categories = ["encoding", "cryptography::cryptocurrencies"]`

src/primitives/gf32.rs
tcharding: ha!

tcharding:
What is the reason we do not implement `PartialOrd` and `Ord`? Is this a finite field math thing?

tcharding:
All suggestions implemented.

apoelstra:
> What is the reason we do not implement `PartialOrd` and `Ord`? Is this a finite field math thing?

Yes. There is no meaningful partial ordering on partial field elements that I can think of. Certainly not a natural one that people would use `<` or `>` to refer to.

apoelstra (approved):
ACK bd010e1f74f8483c8735c5005549894bf51cb7d1

src/primitives/gf32.rs
```diff
@@ -0,0 +1,463 @@
...
+                        Err(Error::InvalidByte(byte))?;
+                    }
+                    Ok(Fe32(byte))
+                }
+            }
+        )+
+    }
+}
+impl_try_from_upper_bounded!(u8 u16 u32 u64 u128);
+impl_try_from_both_bounded!(i8 i16 i32 i64 i128);
```
clarkmoody, line 276: Other than the doc comment, it looks like these implementations are identical.

src/primitives/gf32.rs
```diff
@@ -0,0 +1,463 @@
...
+            Fe32::_9,
+        ]
+        .iter()
+        .copied()
+    }
+
+    /// Creates a field element from a single bech32 character.
+    pub fn from_char(c: char) -> Result<Fe32, Error> {
+        let byte = i8::try_from(u32::from(c)).map_err(|_| Error::InvalidChar(c))?;
+        let byte = byte as u8; // cast guaranteed to be ok since we started with an unsigned value
```
clarkmoody, line 178: Why not cast to `usize` here instead of below?

apoelstra: Good call. I think it we just iterated too many times and wound up with goofy code. Agreed, casting to `usize` directly would be better.

src/primitives/gf32.rs
apoelstra: :) you're right -- we used to have an explicit "negative number" check for the signed variants but we realized that `u8::try_from` would do this check for us and removed it.

So we should combine the macros now.

apoelstra:
cc @tcharding can you fix the two nits that Clark highlighted?

tcharding:
Oh damn, I missed your review last week @clarkmoody, sorry man totally unintentional - I'm on it now!

src/primitives/gf32.rs
tcharding: That was all me, implementing review suggestions wildly without looking close enough at what I was doing :) Thanks, good reviewing.

src/primitives/gf32.rs
tcharding: Changing the cast made me read this code again and think a bit harder. I then simplified it to use explicit value checks instead of using the type conversions to implicitly ensure the array access is within bounds. Next I added a kani verification and fixed the bugs I'd introduced - profit!

tcharding:
Changes in force push:
- Implemented review suggestions
- A few minor docs improvements
- kani verification test

src/primitives/gf32.rs
```diff
@@ -0,0 +1,470 @@
...
+            Fe32::_8,
+            Fe32::_9,
+        ]
+        .iter()
+        .copied()
+    }
+
+    /// Creates a field element from a single bech32 character.
+    pub fn from_char(c: char) -> Result<Fe32, Error> {
+        let ascii = u32::from(c) as usize;
```
apoelstra, line 179: In 45b7cd8:

This cast might truncate. Using `usize::try_from` would be better. Or you could use `i8::try_from`, like the original code did, followed by `usize::try_from`, if you want to be fancy.

src/primitives/gf32.rs
```diff
@@ -0,0 +1,470 @@
...
+    #[test]
+    fn mul_one() {
+        for c in CHARS_LOWER {
+            let fe = Fe32::from_char(c).unwrap();
+            assert_eq!(fe * Fe32::P, fe) // Fe32::P == Fe32(1)
+        }
+    }
+}
+
+#[cfg(kani)]
```
apoelstra, line 456: Nice!

src/primitives/gf32.rs
clarkmoody: Can we trick the test suite into giving us 16-bit `usize`?

clarkmoody (approved):
ACK 45b7cd89772c41222f6c2164e51a1cbed98df99e

src/primitives/gf32.rs
apoelstra: That would be cool, but no, not that I'm aware of.

tcharding: I thought we didn't support 16 bit machines? Want me to add
```rust
// Disable 16-bit support at least for now as we can't guarantee it yet.
#[cfg(target_pointer_width = "16")]
compile_error!(
    "rust-bitcoin currently only supports architectures with pointers wider than 16 bits, let us
    know if you want 16-bit support. Note that we do NOT guarantee that we will implement it!"
);
```

??

tcharding: With `s/rust-bitcoin/rust-bech32/`

src/primitives/gf32.rs
tcharding: Found bugs too, it was mad!

src/primitives/gf32.rs
apoelstra: No, I think we should just support 16-bit machines, since it has zero cost for this crate.

tcharding: I'm not sure this is much better than the original but I've left it as:
```rust
    /// Creates a field element from a single bech32 character.
    pub fn from_char(c: char) -> Result<Fe32, Error> {
        // i8::try_from gets a value in the range 0..=127 since char is unsigned.
        let byte = i8::try_from(u32::from(c)).map_err(|_| Error::InvalidChar(c))?;
        let ascii = usize::try_from(byte).expect("byte value within range 0..=127");
        // We use -1 for any array element that is an invalid char to trigger error from u8::try_from
        let u5 = u8::try_from(CHARS_INV[ascii]).map_err(|_| Error::InvalidChar(c))?;
        Ok(Fe32(u5))
    }
```

apoelstra (approved):
ACK 67dab80cd01c7a3c95fd54d7d84b46eff7bd9e98

apoelstra (approved):
ACK 8a8d33d385838dc8c785e5a48f816ba8a4b72c9d

apoelstra:
@clarkmoody can you do a quick re-ACK? the diff is pretty small between this and the last commit that you did ACK.

apoelstra (approved):
ACK 9f59765ead4b03fae85ca5499f330a7da8af361a

clarkmoody (approved):
ACK 9f59765ead4b03fae85ca5499f330a7da8af361a

Issue #99, opened 2023-04-17
Make check_hrp public


kwek20:
I would like to only verify that the hrp part is valid.
The crate has a method; `check_hrp` but this is private.

I would prefer not to copy the code :) 

Thanks 🙏

tcharding:
We can do that, no worries! Thanks for the issue.

kwek20:
@tcharding Thank you for the work on the new type! What is your release pattern like?

apoelstra:
@kwek20 we are in the middle of a very large refactoring of this crate. I think we'll be done by the end of June, if not sooner, but unfortunately we won't really be in a position to release intermediate crates until it is done (I think).

After that, this crate will probably return to its previous cadence of releasing only as-needed, since it's a small crate which implements pure functions and doesn't need too much upgrading.

tcharding:
Although we should have a pre-release version out soonish.

tcharding:
We can Close this issue now as `Hrp` code was released in the 0.10.0-alpha release.

https://docs.rs/bech32/0.10.0-alpha/bech32/primitives/hrp/struct.Hrp.html

PR #100, opened 2023-04-17
Make check_hrp function part of the public API

tcharding:
As requested by a user of the library; make the `check_hrp` function a public API function.

Requires making the `Case` type public also. Add liberal documentation to `check_hrp` and also to `Case` including links to the relevant bip-173 sections.

Fix: #99

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -469,14 +469,39 @@ impl<U: AsRef<[u8]>> CheckBase32<()> for U {
     }
 }
 
-#[derive(Clone, Copy, PartialEq, Eq)]
-enum Case {
+/// Letter case.
+///
+/// From [BIP-173]: Uppercase/lowercase
+///
+/// > The lowercase form is used when determining a character's value for checksum purposes.
+/// >
+/// > Encoders MUST always output an all lowercase Bech32 string. If an uppercase version of the
+/// > encoding result is desired, (e.g.- for presentation purposes, or QR code use), then an
+/// > uppercasing procedure can be performed external to the encoding process.
+/// >
+/// > Decoders MUST NOT accept strings where some characters are uppercase and some are lowercase (such strings are referred to as mixed case strings).
+/// >
+/// > For presentation, lowercase is usually preferable, but inside QR codes uppercase SHOULD be
+/// > used, as those permit the use of alphanumeric mode, which is 45% more compact than the normal
+/// > byte mode.
+///
+/// [BIP173]: <https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#user-content-Bech32>
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum Case {
+    /// Uppercase letters.
     Upper,
+    /// Lowercase letters.
     Lower,
+    /// No letter case, i.e., all characters were digits.
     None,
 }
 
-/// Checks if the human-readable part (HRP) is valid.
+/// Checks if the human-readable part (HRP) is valid as specified by [BIP-173].
+///
+/// > The human-readable part, which is intended to convey the type of data, or anything else that
+/// > is relevant to the reader. This part MUST contain 1 to 83 US-ASCII characters, with each
+/// > character having a value in the range [33-126]. HRP validity may be further restricted by
+/// > specific applications.
 ///
 /// # Returns
 ///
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -469,14 +469,39 @@ impl<U: AsRef<[u8]>> CheckBase32<()> for U {
...
+///
+/// [BIP173]: <https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#user-content-Bech32>
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum Case {
+    /// Uppercase letters.
     Upper,
+    /// Lowercase letters.
     Lower,
+    /// No letter case, i.e., all characters were digits.
     None,
```
apoelstra, line 496: I don't like this variant and think it should be folded into `Lower`.  AFAIK there is no HRP defined by anybody which is all digits, so any codepaths that involve this variant will be unused.

I think that `check_hrp` should just return `()` or an error. If people want to check case-consistency of the whole string they should use `check_characters` and check this at the same time that they're checking that every non-HRP character is a valid bech32 string.

Maaybe it's marginally more efficient to do things this way, but adding a whole extra type to the API, which is an enum where 1/3 of the variants will literally never be used, is a pretty big cost.

clarkmoody:
We're doing lots of manual matching on the results of `check_hrp`. What if we introduced a new `Hrp` type and used the "Parse, Don't Validate" methodology?

```rust
pub enum Hrp { // Or `HumanReadablePart` ?
    Lower(&str),
    Upper(&str),
}

impl Hrp {
    // This replaces the current `check_hrp` function
    pub fn parse(s: &str) -> Result<Self, Error> { .. }

    // When we need the lowercase representation
    pub fn to_lower(&self) -> Cow<&str> {
        match check_hrp(hrp)? {
            Self::Upper(upper) => Cow::Owned(upper.to_lowercase()),
            Self::Lower(lower) => Cow::Borrowed(lower),
        }
    }
}

// Implement this as well
impl TryFrom<&str> for Hrp {
  // Calls Hrp::parse under the hood
}
```

This would change the API elsewhere as well. So instead of `hrp: &str`, we would have `hrp: Hrp`, and the caller would be in charge of supplying a parsed `Hrp` instance. This simplifies the internals of a few of our functions because it is no longer possible for the user to supply an invalid HRP.

Thoughts?

apoelstra:
I like this idea, but we still need to think about how to handle the all-numeric HRP case. Maybe we should just panic? Annoyingly BIP 173 explicitly allows this case (and even has a test vector for it).

I suppose we could just add a third `Numeric(&str)` variant to the `Hrp` type. I think this would be fine because it's much less likely for users to match on the new `Hrp` type than on the existing `Case` type.

tcharding:
Today I was working on the re-write branch (`04-04-primitives-api`) and I totally removed the `Case` type. The bips explicitly state that implementations should only ever return lowercase encodings so the only case stuff is checking that there is no mixed case. It made implementing all the test vectors harder but made the lib way better.

Thanks for the reviews, perhaps I went off half cocked (I got excited that a user came along with a request :)

Perhaps we should put this on ice till we make more progress with the re-write?

apoelstra:
@tcharding we don't have `Cow` without an allocator, so making `check_hrp` return one means it's impossible to do anything at all without an allocator.

Previously you only needed an allocator to get the lowercased version of an HRP, which has pretty limited use.

apoelstra:
Regarding "smells like C code", I guess you mean because you're iterating over the bytes of a string and using `| 32` to lowercase it. This is fine. Rust has basically no support for ASCII because its stdlib is mostly targeted at web developers who think "systems language" sounds cool.

People who actually do systems programming are so fucked up from C that they'll tolerate anything, so there's not a lot of value in supporting them anyway, I'd guess..

In any case, I think the `|32` stuff and the `string.bytes().enumerate()` stuff is fine.

The `cfg(feature = "alloc")` stuff still seems overdone ... I really think don't think we should be requiring an allocator anywhere except some convenience functions at the edge of the API ... but we can iterate on that when we get closer to PR'ing this stuff.

tcharding:
> Rust has basically no support for ASCII because its stdlib is mostly targeted at web developers who think "systems language" sounds cool.

LOL, explains why I don't enjoy Rust meet ups, I was wondering why there were so many web devs there.

apoelstra:
It might be interesting to try to find/create "Rust for Linux kernel development" type meetups -- I suspect that kernel development has similar constraints to us (lots of simple ASCII encodings, aversion to async/await, need for precise allocation control, distrust of cargo and crates.io, etc etc). And will probably be full of people closer to our age :P.

tcharding:
Your comment made me pull down the kernel tree and have a look at what they had been up to. We probably can garner a few tips and ideas from there.

tcharding:
Superseeded by #103

PR #101, opened 2023-04-22
use new `Fe32` type in place of `gf32`

apoelstra:
This requires dropping `Ord` impls from the error types, but IMHO these didn't really make sense anyway.

This also keeps the `u5` name. We will rename this in a later PR which replaces the *Base32 traits with iterator adaptors, since at that point there will be a much smaller API surface to change.

Followup to #98.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -59,78 +59,7 @@ pub mod primitives;
 
 #[cfg(feature = "arrayvec")]
 use arrayvec::{ArrayVec, CapacityError};
-
-/// Integer in the range `0..32`.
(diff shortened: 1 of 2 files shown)


tcharding (approved):
ACK 22f20523de53e2baa71ccbd32c1efe1330284a3f

tcharding:
Sweet, a much more simple next step than I came up with when playing around, nice one.

apoelstra:
cc @clarkmoody can you take a look at this (or ok us to merge it without you?)

clarkmoody (approved):
Changes LGTM :shipit:

PR #102, opened 2023-04-23
Rewrite bech32 crate

tcharding:
Draft because done on top of #101, also because @apoelstra suggested doing this in smaller parts but I think it might be easier to just pull the trigger and do it all at once. @clarkmoody please say if you find this approach too much and would prefer to see the PR broken up into multiple PRs.

### Why the rewrite?

- It was observed that the writer API could be generalized to use an iterator API.
- It was observed that the checksum logic is an implementation of BCH codes, this is the only cryptography part of the crate, as such it is complex, isolating it helps reduce complexity of the code for non-cryptographers.
- ~This is a bitcoin crate but currently it supports arbitrary HRP prefixes, by making the top level API more bitcoin specific we can simplify the API.~

### Aims

- Make a `segwit` API that is ergonomic to use for Bitcoin users and that implements the specification laid out in BIP-173 And BIP-350
- Isolate the cryptography to a submodule
- Create a stable API that we can commit to maintaining

For example usage in `rust-bitcoin` see: https://github.com/rust-bitcoin/rust-bitcoin/pull/1809


src/lib.rs
```diff
@@ -45,869 +57,455 @@ extern crate alloc;
...
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum Bech32 {}
+
+/// The bech32m checksum algorithm, defined in BIP-350.
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum Bech32m {}
+
+/// The codex32 checksum algorithm, defined in BIP-93.
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+enum Codex32 {}
```
tcharding, line 440: Do we want to keep this in?

tcharding:
Notes:

- The fuzzing needs fixing
- I've managed to loose the benchmark code you wrote @apoelstra

src/primitives/hrpstring.rs
```diff
@@ -0,0 +1,558 @@
...
+    fn try_from_err() {
+        assert!(Fe32::try_from(32_u8).is_err());
+        assert!(Fe32::try_from(32_u16).is_err());
+        assert!(Fe32::try_from(32_u32).is_err());
+        assert!(Fe32::try_from(32_u64).is_err());
+        assert!(Fe32::try_from(32_u128).is_err());
+    }
+
+    #[test]
+    #[allow(unused_variables)] // Triggered by matches macro.
```
apoelstra, line 445: This is triggered because the `matches` macro is used incorrectly.

src/primitives/gf1024.rs
```diff
@@ -0,0 +1,14 @@
+// Written by Clark Moody and the rust-bitcoin developers.
+// SPDX-License-Identifier: MIT
+
+//! GF1024 - Galois Field over 1024 elements.
```
apoelstra, line 4: I'm unclear on the purpose of including this in this PR.

src/lib.rs
```diff
@@ -45,869 +57,454 @@ extern crate alloc;
...
+        0x1fbd920fffe7bee52,
+        0x1739640bdeee3fdad,
+        0x07729a039cfc75f5a,
+    ];
+    const TARGET_RESIDUE: u128 = 0x10ce0795c2fd1e62a;
+}
+
+/// True if program length is either 20 bytes (p2wpkh) or 32 bytes (p2wsh).
+/// ref: [BIP-141](https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#user-content-Witness_program)
+#[cfg(feature = "alloc")]
```
apoelstra, line 481: what part of this function requires an allocator?

src/primitives/gf1024.rs
tcharding: Will drop, thanks.

src/primitives/hrpstring.rs
tcharding: Removed, and fixed tests to fully implement the test vectors from both bips.

src/lib.rs
tcharding: This is to prevent clippy warning when "alloc" is not enabled, I added a comment to it to say as such.

tcharding:
Is it even worth fuzzing this crate? We accept random data and convert it to a string (filters out invalid utf-8) and then decode (filters out invalid bech32  characters). Won't the number of times this actually succeeds be so small that running the fuzzing is essentially doing nothing? Or is the fuzzing meaningfully helping to prove `1.` below?

Am I correct in thinking we want to test that:
1. arbitrary `&str` doesn't cause us to panic
2. all valid bech32 strings are correctly parsed and roundtripped

apoelstra:
You can test whether it's able to produce valid checksums. It would surprise me if it can't, given how simple the algorithm is and the fact that it only requires modiying the last 6 characters of the string (and modifying them independently, by simply xoring them with the correct value).

But regardless, it is valuable to fuzz that we don't panic.

src/lib.rs
apoelstra: No, we'll drop it from the final PR.

src/network.rs
```diff
@@ -0,0 +1,75 @@
...
+    ///
+    /// ```
+    /// # use bech32::Network;
+    /// assert_eq!(Network::Bitcoin.to_hrp(), "bc");
+    /// assert_eq!(Network::Testnet.to_hrp(), "tb");
+    /// assert_eq!(Network::Regtest.to_hrp(), "bcrt");
+    /// ```
+    ///
+    /// [segwit address format]: <https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#user-content-Segwit_address_format>
+    pub const fn to_hrp(&self) -> &'static str {
```
clarkmoody, line 41: Why not return an instance of the human-readable part type instead of a string slice?

tcharding: I'm having a play with introducing the `Hrp` type as you suggested previously, cheers.

src/primitives/hrpstring.rs
```diff
@@ -0,0 +1,460 @@
...
+use crate::witness_version::{self, WitnessVersion};
+
+/// Maximum length of the human-readable part, as defined by BIP-173.
+const MAX_HRP_LEN: usize = 83;
+
+/// An HRP string that has been parsed from an ordinary checksummed string.
+///
+/// Parsing as an HRP string does not validate the checksum in any way.
+#[derive(Debug)]
+pub struct Parsed<'s> {
```
clarkmoody, line 36: The names here don't make a ton of sense. The module is called `hrpstring`, but the primary data structure is `Parsed` and contains not only HRP data but also witness version and the data part.

apoelstra: The data structure is an HRP string which has been parsed into an HRP, witness version, and data part.

We should change this if it's not understandable but it seemed pretty straightforward to me.

src/network.rs
tcharding: Aight, I hacked up a new version of this PR that adds an `Hrp` type, and also a `KnownHrp` type for segwit specific code. Could be mistakes because it was a fair bit of work and I should really go over it on another day before pushing but I'm pretty psyched for this so if you'll excuse me I'll push right away.

tcharding:
The major change here is that the PR no longer removes the current API. Instead it adds a `segwit` module that implements bip-173 and 350 using the new `primitives` stuff. Also:

- Adds `Hrp` type
- Adds `KnownHrp` type
- Uses new proposed `rust-bitcoin-primitives` crate to get the `WitnessVersion`

Thanks for your patience @clarkmoody and keeping on prodding me about the `Hrp` type, this worked out really nicely.

tcharding:
Closing since this is stale.

PR #103, opened 2023-04-28
Add `Hrp` type

tcharding:
Instead of accepting `&str` for hrp strings we can add a type `Hrp` and abstract the checking of the string. This is the "parse don't validate" methodology suggested by clarkmoody [here](https://github.com/rust-bitcoin/rust-bech32/pull/100#issuecomment-1512435732).

This is an invasive breaking change.

### DIFF
--- a/fuzz/fuzz_targets/encode_decode.rs
+++ b/fuzz/fuzz_targets/encode_decode.rs
@@ -1,6 +1,9 @@
 extern crate bech32;
 
 use std::convert::TryFrom;
+use std::str;
+
+use bech32::Hrp;
 
 fn do_test(data: &[u8]) {
     if data.len() < 1 {
@@ -13,10 +16,6 @@ fn do_test(data: &[u8]) {
         return;
     }
 
-    let hrp = String::from_utf8_lossy(&data[1..hrp_end])
-        .to_lowercase()
-        .to_string();
-
     let dp = data[hrp_end..]
         .iter()
         .map(|b| bech32::u5::try_from(b % 32).unwrap())
@@ -28,11 +27,21 @@ fn do_test(data: &[u8]) {
         bech32::Variant::Bech32
     };
 
-    if let Ok(data_str) = bech32::encode(&hrp, &dp, variant).map(|b32| b32.to_string()) {
-        let decoded = bech32::decode(&data_str);
-        let b32 = decoded.expect("should be able to decode own encoding");
+    match str::from_utf8(&data[1..hrp_end]) {
+        Err(_) => return,
+        Ok(s) => {
+            match Hrp::parse(&s) {
+                Err(_) => return,
+                Ok(hrp) => {
+                    if let Ok(data_str) = bech32::encode(hrp, &dp, variant).map(|b32| b32.to_string()) {
+                        let decoded = bech32::decode(&data_str);
+                        let b32 = decoded.expect("should be able to decode own encoding");
 
-        assert_eq!(bech32::encode(&b32.0, &b32.1, b32.2).unwrap(), data_str);
+                        assert_eq!(bech32::encode(b32.0, &b32.1, b32.2).unwrap(), data_str);
+                    }
+                }
+            }
+        }
     }
 }
 
--- a/fuzz/fuzz_targets/parse_hrp.rs
+++ b/fuzz/fuzz_targets/parse_hrp.rs
@@ -1,14 +1,13 @@
 extern crate bech32;
 
-fn do_test(data: &[u8]) {
-    let data_str = String::from_utf8_lossy(data);
-    let decoded = bech32::decode(&data_str);
-    let b32 = match decoded {
-        Ok(b32) => b32,
-        Err(_) => return,
-    };
+use bech32::Hrp;
 
-    assert_eq!(bech32::encode(b32.0, b32.1, b32.2).unwrap(), data_str);
+fn do_test(data: &[u8]) {
+    let s = String::from_utf8_lossy(data);
+
+    // Make sure parsing garbage doesn't make us crash (from_utf8_lossy should
+    // contain some garbage, perhaps even invalid chars).
+    let _ = Hrp::parse(&s);
 }
 
 #[cfg(feature = "afl")]
@@ -36,7 +35,7 @@ fn main() {
 mod tests {
     fn extend_vec_from_hex(hex: &str, out: &mut Vec<u8>) {
         let mut b = 0;
-        for (idx, c) in hex.as_bytes().iter().enumerate() {
+        for (idx, c) in hex.as_bytes().iter().filter(|&&c| c != b'\n').enumerate() {
             b <<= 4;
             match *c {
                 b'A'...b'F' => b |= c - b'A' + 10,
@@ -54,7 +53,7 @@ mod tests {
     #[test]
     fn duplicate_crash() {
         let mut a = Vec::new();
-        extend_vec_from_hex("00000000", &mut a);
+        extend_vec_from_hex("ff6c2d", &mut a);
         super::do_test(&a);
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,11 +17,12 @@
     doc = "
 # Examples
 ```
-use bech32::{self, FromBase32, ToBase32, Variant};
-let encoded = bech32::encode(\"bech32\", vec![0x00, 0x01, 0x02].to_base32(), Variant::Bech32).unwrap();
+use bech32::{self, FromBase32, ToBase32, Variant, Hrp};
+let hrp = Hrp::parse(\"bech32\").expect(\"bech32 is valid\");
+let encoded = bech32::encode(hrp, vec![0x00, 0x01, 0x02].to_base32(), Variant::Bech32).unwrap();
 assert_eq!(encoded, \"bech321qqqsyrhqy2a\".to_string());
 let (hrp, data, variant) = bech32::decode(&encoded).unwrap();
-assert_eq!(hrp, \"bech32\");
+assert_eq!(hrp.to_string(), \"bech32\");
 assert_eq!(Vec::<u8>::from_base32(&data).unwrap(), vec![0x00, 0x01, 0x02]);
 assert_eq!(variant, Variant::Bech32);
 ```
@@ -46,13 +47,12 @@ extern crate alloc;
 extern crate core;
 
 #[cfg(all(feature = "alloc", not(feature = "std"), not(test)))]
-use alloc::borrow::Cow;
-#[cfg(all(feature = "alloc", not(feature = "std"), not(test)))]
 use alloc::{string::String, vec::Vec};
 use core::convert::{Infallible, TryFrom};
 use core::{fmt, mem};
-#[cfg(any(feature = "std", all(feature = "alloc", test)))]
-use std::borrow::Cow;
+
+use crate::primitives::hrp;
+pub use crate::primitives::hrp::Hrp;
 
 mod error;
 pub mod primitives;
@@ -113,21 +113,23 @@ impl<'a> Bech32Writer<'a> {
     /// This is a rather low-level API and doesn't check the HRP or data length for standard
     /// compliance.
     pub fn new(
-        hrp: &str,
+        hrp: Hrp,
         variant: Variant,
         fmt: &'a mut dyn fmt::Write,
     ) -> Result<Bech32Writer<'a>, fmt::Error> {
         let mut writer = Bech32Writer { formatter: fmt, chk: 1, variant };
 
-        writer.formatter.write_str(hrp)?;
+        for c in hrp.lowercase_char_iter() {
+            writer.formatter.write_char(c)?;
+        }
         writer.formatter.write_char(SEP)?;
 
         // expand HRP
-        for b in hrp.bytes() {
+        for b in hrp.lowercase_byte_iter() {
             writer.polymod_step(u5(b >> 5));
         }
         writer.polymod_step(u5(0));
-        for b in hrp.bytes() {
+        for b in hrp.lowercase_byte_iter() {
             writer.polymod_step(u5(b & 0x1f));
         }
 
@@ -235,7 +237,7 @@ write_base_n! { WriteBase32, u5, write_u5 }
 write_base_n! { WriteBase256, u8, write_u8 }
 
 #[cfg(feature = "arrayvec")]
-#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 /// Combination of Errors for use with array vec
 pub enum ComboError {
     /// Error from this crate
@@ -250,6 +252,10 @@ impl From<Error> for ComboError {
 #[cfg(feature = "arrayvec")]
 impl From<CapacityError> for ComboError {
     fn from(e: CapacityError) -> ComboError { ComboError::WriteError(e) }
+}
+#[cfg(feature = "arrayvec")]
+impl From<hrp::Error> for ComboError {
+    fn from(e: hrp::Error) -> ComboError { ComboError::Bech32Error(Error::Hrp(e)) }
 }
 
 #[cfg(feature = "arrayvec")]
@@ -399,78 +405,30 @@ impl<U: AsRef<[u8]>> CheckBase32<()> for U {
     }
 }
 
-#[derive(Clone, Copy, PartialEq, Eq)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
 enum Case {
     Upper,
     Lower,
     None,
 }
 
-/// Checks if the human-readable part (HRP) is valid.
-///
-/// # Returns
-///
-/// The case of the HRP, if HRP is valid.
-///
-/// # Errors
-///
-/// * **MixedCase**: If the HRP contains both uppercase and lowercase characters.
-/// * **InvalidChar**: If the HRP contains any non-ASCII characters (outside 33..=126).
-/// * **InvalidLength**: If the HRP is outside 1..83 characters long.
-fn check_hrp(hrp: &str) -> Result<Case, Error> {
-    if hrp.is_empty() || hrp.len() > 83 {
-        return Err(Error::InvalidLength);
-    }
-
-    let mut has_lower: bool = false;
-    let mut has_upper: bool = false;
-    for b in hrp.bytes() {
-        // Valid subset of ASCII
-        if !(33..=126).contains(&b) {
-            return Err(Error::InvalidChar(b as char));
-        }
-
-        if b.is_ascii_lowercase() {
-            has_lower = true;
-        } else if b.is_ascii_uppercase() {
-            has_upper = true;
-        };
-
-        if has_lower && has_upper {
-            return Err(Error::MixedCase);
-        }
-    }
-
-    Ok(match (has_upper, has_lower) {
-        (true, false) => Case::Upper,
-        (false, true) => Case::Lower,
-        (false, false) => Case::None,
-        (true, true) => unreachable!(),
-    })
-}
-
-/// Encodes a bech32 payload to a writer ([`fmt::Write`]).
+/// Encodes a bech32 payload to a writer ([`fmt::Write`]) using lowercase.
 ///
 /// This method is intended for implementing traits from [`std::fmt`].
 ///
 /// # Errors
 ///
-/// * If human-readable part (`hrp`) is invalid.
 /// * Deviations from standard.
 /// * No length limits are enforced for the data part.
-#[cfg(feature = "alloc")]
 pub fn encode_to_fmt<T: AsRef<[u5]>>(
     fmt: &mut dyn fmt::Write,
-    hrp: &str,
+    hrp: Hrp,
     data: T,
     variant: Variant,
 ) -> Result<fmt::Result, Error> {
-    let hrp_lower = match check_hrp(hrp)? {
-        Case::Upper => Cow::Owned(hrp.to_lowercase()),
-        Case::Lower | Case::None => Cow::Borrowed(hrp),
-    };
-
-    encode_to_fmt_anycase(fmt, &hrp_lower, data, variant)
+    let mut hrp = hrp;
+    hrp.lowercase();
+    encode_to_fmt_anycase(fmt, hrp, data, variant)
 }
 
 /// Encode a bech32 payload to an [fmt::Write], but with any case.
@@ -479,7 +437,7 @@ pub fn encode_to_fmt<T: AsRef<[u5]>>(
 /// See `encode_to_fmt` for meaning of errors.
 pub fn encode_to_fmt_anycase<T: AsRef<[u5]>>(
     fmt: &mut dyn fmt::Write,
-    hrp: &str,
+    hrp: Hrp,
     data: T,
     variant: Variant,
 ) -> Result<fmt::Result, Error> {
@@ -500,22 +458,17 @@ pub fn encode_to_fmt_anycase<T: AsRef<[u5]>>(
 ///
 /// # Errors
 ///
-/// * If human-readable part (`hrp`) is invalid.
 /// * Deviations from standard.
 /// * No length limits are enforced for the data part.
 pub fn encode_without_checksum_to_fmt<T: AsRef<[u5]>>(
     fmt: &mut dyn fmt::Write,
-    hrp: &str,
+    hrp: Hrp,
     data: T,
 ) -> Result<fmt::Result, Error> {
-    #[cfg(feature = "alloc")]
-    let hrp = match check_hrp(hrp)? {
-        Case::Upper => Cow::Owned(hrp.to_lowercase()),
-        Case::Lower | Case::None => Cow::Borrowed(hrp),
-    };
-
-    if let Err(e) = fmt.write_str(&hrp) {
-        return Ok(Err(e));
+    for c in hrp.lowercase_char_iter() {
+        if let Err(e) = fmt.write_char(c) {
+            return Ok(Err(e));
+        }
     }
     if let Err(e) = fmt.write_char(SEP) {
         return Ok(Err(e));
@@ -562,11 +515,10 @@ impl Variant {
 ///
 /// # Errors
 ///
-/// * If human-readable part (`hrp`) is invalid.
 /// * Deviations from standard.
 /// * No length limits are enforced for the data part.
 #[cfg(feature = "alloc")]
-pub fn encode<T: AsRef<[u5]>>(hrp: &str, data: T, variant: Variant) -> Result<String, Error> {
+pub fn encode<T: AsRef<[u5]>>(hrp: Hrp, data: T, variant: Variant) -> Result<String, Error> {
     let mut buf = String::new();
     encode_to_fmt(&mut buf, hrp, data, variant)?.unwrap();
     Ok(buf)
@@ -576,11 +528,10 @@ pub fn encode<T: AsRef<[u5]>>(hrp: &str, data: T, variant: Variant) -> Result<St
 ///
 /// # Errors
 ///
-/// * If human-readable part (`hrp`) is invalid.
 /// * Deviations from standard.
 /// * No length limits are enforced for the data part.
 #[cfg(feature = "alloc")]
-pub fn encode_without_checksum<T: AsRef<[u5]>>(hrp: &str, data: T) -> Result<String, Error> {
+pub fn encode_without_checksum<T: AsRef<[u5]>>(hrp: Hrp, data: T) -> Result<String, Error> {
     let mut buf = String::new();
     encode_without_checksum_to_fmt(&mut buf, hrp, data)?.unwrap();
     Ok(buf)
@@ -592,14 +543,14 @@ pub fn encode_without_checksum<T: AsRef<[u5]>>(hrp: &str, data: T) -> Result<Str
 ///
 /// The human-readable part in lowercase, the data with the checksum removed, and the encoding.
 #[cfg(feature = "alloc")]
-pub fn decode(s: &str) -> Result<(String, Vec<u5>, Variant), Error> {
+pub fn decode(s: &str) -> Result<(Hrp, Vec<u5>, Variant), Error> {
     let (hrp_lower, mut data) = split_and_decode(s)?;
     if data.len() < CHECKSUM_LENGTH {
         return Err(Error::InvalidLength);
     }
 
     // Ensure checksum
-    match verify_checksum(hrp_lower.as_bytes(), &data) {
+    match verify_checksum(hrp_lower, &data) {
         Some(variant) => {
             // Remove checksum from data payload
             data.truncate(data.len() - CHECKSUM_LENGTH);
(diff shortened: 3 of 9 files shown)


tcharding:
Please note, I'm not happy with the inefficient way this PR converts an `Hrp` to a lowercase version in `encode_to_fmt`.

src/primitives/hrp.rs
```diff
@@ -0,0 +1,180 @@
...
+        }
+    }
+
+    /// Creates a lowercase iterator over the bytes (ASCII characters) of this HRP.
+    pub fn iter(&self) -> LowercaseIter<'s> {
+        let s = match self {
+            Self::Upper(s) => s,
+            Self::Lower(s) => s,
+            Self::Digit(s) => s,
+        };
```
clarkmoody, line 95: Reuse `self.as_str` here and in `len` below?

tcharding: Works for length but I get lifetime errors here. Thanks! I'll give you time to finish review before I force push.

clarkmoody: This seems to work for me (put `'s` lifetime on `self`)

```rust
    pub fn iter(&'s self) -> LowercaseIter<'s> {
        let iter = self.as_str().bytes();
        LowercaseIter { iter }
    }
```

tcharding: haha thanks man, you can code for me anytime :)

tcharding: Even better, we can just elide the lifetime on `LowercaseIter<'_>`

clarkmoody: 👌

src/lib.rs
```diff
@@ -722,13 +674,11 @@ fn split_and_decode(s: &str) -> Result<(String, Vec<u5>), Error> {
             if c.is_lowercase() {
                 match case {
                     Case::Upper => return Err(Error::MixedCase),
-                    Case::None => case = Case::Lower,
```
apoelstra, line 725: In e0bfa91dc74888005246ae80b46d0a620827888f:

Removing these `Case::None` cases introduces a bug in case the HRP is all digits but the data part has uppercase letters.

src/primitives/hrp.rs
```diff
@@ -0,0 +1,180 @@
...
+
+/// The human-readable part (human readable prefix before the '1' separator).
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum Hrp<'a> {
+    /// Lowercase human-readable part.
+    Lower(&'a str),
+    /// Uppercase human-readable part.
+    Upper(&'a str),
+    /// All digit human-readable part.
+    Digit(&'a str),
```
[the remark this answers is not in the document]
apoelstra: I guess we will need it if we want to fix the case-checking bug in the top-level parsing function.... but I don't think we should bother, I think we should have a `check_characters` function that does validity checks on the whole string rather than checking for consistency with the HRP.

Once we do that I'm not sure that there's any value in having distinct `Lower`/`Upper` variants at all...we should just unconditionally use bitshifts to convert case when it's needed.

src/primitives/hrp.rs
```diff
@@ -0,0 +1,180 @@
...
+        Ok(Hrp::Digit(hrp))
+    }
+
+    /// Returns the human-readable part as a lowercase string.
+    #[cfg(feature = "alloc")]
+    pub fn to_lower(&self) -> Cow<str> {
+        match self {
+            Self::Upper(upper) => Cow::Owned(upper.to_lowercase()),
+            Self::Lower(lower) => Cow::Borrowed(lower),
+            Self::Digit(digits) => Cow::Borrowed(digits),
```
apoelstra, line 85: I guess this would be less efficient if we weren't tracking case, but I don't really see the use of this method anyway, so I think that's fine.

src/primitives/hrp.rs
```diff
@@ -0,0 +1,180 @@
...
+#[cfg(feature = "alloc")]
+use alloc::borrow::Cow;
+use core::fmt;
+use core::iter::FusedIterator;
+
+/// Maximum length of the human-readable part, as defined by BIP-173.
+const MAX_HRP_LEN: usize = 83;
+
+/// The human-readable part (human readable prefix before the '1' separator).
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
```
apoelstra, line 22: In e0bfa91dc74888005246ae80b46d0a620827888f:

I think we should do a case-insensitive comparison rather than deriving PartialEq/Eq.

apoelstra: And I think we should impl `Ord` similarly.

src/lib.rs
```diff
@@ -474,74 +477,32 @@ impl<U: AsRef<[u8]>> CheckBase32<()> for U {
...
-/// This method is intended for implementing traits from [`std::fmt`].
+/// This method is intended for implementing traits from [`std::fmt`]. Please note that if `hrp` is
+/// uppercase this function will re-parse it to get a lowercase version, this could be slow.
 ///
 /// # Errors
 ///
-/// * If human-readable part (`hrp`) is invalid.
 /// * Deviations from standard.
 /// * No length limits are enforced for the data part.
 #[cfg(feature = "alloc")]
```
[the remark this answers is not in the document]
tcharding: This is fixed now, no alloc feature needed for `encode_to_fmt`.

src/lib.rs
tcharding: Added this test, does this cover what you meant?
```rust
    
    #[test]
    #[cfg(feature = "alloc")]
    fn attempt_decode_all_digit_hrp_mixed_case_data() {
        let addr = "23451qar0sRRR7xfkvy5l643lydnw9re59gtzzwf5mdq";
        assert!(crate::decode(addr).is_err());
    }
```

tcharding: I believe this is ok now, I added a test to verify it.
```rust
    #[test]
    #[cfg(feature = "alloc")]
    fn decode_all_digit_hrp_uppercase_data() {
        let addr = "23451QAR0SRRR7XFKVY5L643LYDNW9RE59GTZZLKULZK";
        let (hrp, data, variant) = crate::decode(&addr).expect("address is well formed");
        assert_eq!(hrp, Hrp::parse("2345").unwrap().0);
        let (hrp, _) = Hrp::parse("2345").unwrap();
        let s = crate::encode(hrp, &data, variant).expect("failed to encode");
        assert_eq!(s, addr);
    }
```

tcharding: I added `Case::None` back in and added this test
```rust  
    #[test]
    #[cfg(feature = "alloc")]
    fn decode_all_digit_hrp_uppercase_data() {
        let addr = "23451QAR0SRRR7XFKVY5L643LYDNW9RE59GTZZLKULZK";
        let (hrp, data, variant) = crate::decode(&addr).expect("address is well formed");
        assert_eq!(hrp, Hrp::parse("2345").unwrap());
        let hrp = Hrp::parse("2345").unwrap();
        let s = crate::encode(hrp, &data, variant).expect("failed to encode");
        assert_eq!(s.to_uppercase(), addr);
    }
```

src/primitives/hrp.rs
tcharding: Done!

tcharding:
Changes in force push are substantial, `Hrp` now has an 83 byte long buffer which stores the string as ascii byte values. We provide iterators for getting bytes, chars, lowercase bytes, and lowercase chars.

Please note the logic in `split_and_decode` is changed, the returned `Hrp` has not been lowercased like the old string that was returned.

tcharding:
Converted to draft so we do not merge because `./fuzz.sh encode_decode` crashes on my machine with the following hex:
045f5f5f5f5f5f5f5f0000000000000000083535
ffa8531a004b3e0000b1f15ab026bd0d4fc2eb72825f7cf1570000000000
0000040000000000

Two interesting things:
1. I cannot reproduce the failures, this test passes

```rust

    #[test]
    fn duplicate_crash() {
        let mut a = Vec::new();
        extend_vec_from_hex("045f5f5f5f5f5f5f5f0000000000000000083535", &mut a);
        super::do_test(&a);

        let mut a = Vec::new();
        extend_vec_from_hex("ffa8531a004b3e0000b1f15ab026bd0d4fc2eb72825f7cf1570000000000", &mut a);
        super::do_test(&a);

        let mut a = Vec::new();
        extend_vec_from_hex("0000040000000000", &mut a);
        super::do_test(&a);
    }
```

2. The fuzzer always fails with the same values so it must be starting from the same initial state, this is what we were wondering about @apoelstra.

apoelstra:
@tcharding can you reproduce the failures if you test with `RUSTFLAGS=--cfg=fuzzing`?

If so you have been burned by the kind of bullshit that https://github.com/rust-bitcoin/rust-bitcoin/pull/1821 will fix.

apoelstra:
> 2\. The fuzzer always fails with the same values so it must be starting from the same initial state, this is what we were wondering about @apoelstra.

This does not sound like a valid inference.

tcharding:
> > 2. The fuzzer always fails with the same values so it must be starting from the same initial state, this is what we were wondering about @apoelstra.
> 
> This does not sound like a valid inference.

So I guessed from your comment that I don't understand how fuzzers work properly. I did some more reading. From my understanding the fuzzer starts with some input, runs the test, measures the coverage, then mutates the input in some interesting way and repeats. I thought that the initial input was random, hence why I was surprised that it failed with the same values on multiple runs.

Where is the hole in my understanding?

tcharding:
> @tcharding can you reproduce the failures if you test with `RUSTFLAGS=--cfg=fuzzing`?
> 
> If so you have been burned by the kind of bullshit that [rust-bitcoin/rust-bitcoin#1821](https://github.com/rust-bitcoin/rust-bitcoin/pull/1821) will fix.

I can't repro. Had a go this morning, left it a while, and had another go but still no dice?

apoelstra:
@tcharding try deleting your `hfuzz_workspace` directory then try again.

>Where is the hole in my understanding?

There are many paths to the same state.

But also, in this case, I think you're confused by the hex output from `./travis-fuzz.sh` which just looks in the `hfuzz_workspace` directory for the first crash file ... so if you don't delete that between runs, it'll output the same file. (And if you fix the bug, it may confusingly still be output.)

tcharding:
ooooh, would it be useful to add at the start of the loop (to all the fuzzing scripts):
```bash

# Clean up crash file from previous runs.
if [ -f hfuzz_workspace/$FILE/HONGGFUZZ.REPORT.TXT ]; then
        rm hfuzz_workspace/$FILE/HONGGFUZZ.REPORT.TXT
fi
```

tcharding:
Fixed up a few docs mistakes and removed infrastructure changes to fuzz files. @apoelstra please review carefully the changes in `encode_decode` - I feel like I'm still not groking fuzzing, I'm unsure if the code I've written is the right way to write fuzzing code.

tcharding:
Changes in force push, I found a bug while using this in `bitcoin`. Added a test that catches it and fixed it. (I wasn't using the lowercase iterator when writing to polymod function). Was not being caught because we lowercase in the API functions higher up.

fuzz/fuzz_targets/encode_decode.rs
```diff
@@ -2,20 +2,33 @@ extern crate bech32;
...
+
 fn do_test(data: &[u8]) {
     if data.len() < 1 {
+        println!("early return 0");
         return;
     }
 
-    let hrp_end = (data[0] as usize) + 1;
+    // Get a random valid length hrp 1-83 bytes.
+    let hrp_end = ((data[0] as usize) % 82) + 1;
```
apoelstra, line 14: In 599ce6d089904386ac6a34d77a6b3b1fe222393f:

This actually only gives numbers in the range 1-82.

fuzz/fuzz_targets/encode_decode.rs
```diff
@@ -2,20 +2,33 @@ extern crate bech32;
...
-        .to_string();
+    let mut hrp = [0_u8; 83];
+    for i in 0..hrp_end {
+        // valid ascii range of hrp characters is 33-126
+        hrp[i] = (data[i] % 94) + 33;
+    }
+
+    let hrp = hrp.iter().take(hrp_end).map(|b| unsafe {
+        char::from_u32_unchecked(*b as u32)
+    }).collect::<String>();
```
apoelstra, line 30: In 599ce6d089904386ac6a34d77a6b3b1fe222393f:

Why is this safe?

apoelstra: Oh, I see, you're translating the data in the lines above. I wouldn't do that.

I'd just take the first byte as a "length", try to parse up to that many bytes as an HRP, and use the rest as bech32 data.

fuzz/fuzz_targets/encode_decode.rs
```diff
@@ -28,11 +41,13 @@ fn do_test(data: &[u8]) {
         bech32::Variant::Bech32
     };
 
-    if let Ok(data_str) = bech32::encode(&hrp, &dp, variant).map(|b32| b32.to_string()) {
-        let decoded = bech32::decode(&data_str);
-        let b32 = decoded.expect("should be able to decode own encoding");
+    if let Ok(hrp) = Hrp::parse(&hrp) {
+        if let Ok(encoded) = bech32::encode(hrp, &dp, variant).map(|b32| b32.to_string()) {
```
apoelstra, line 45: In 599ce6d089904386ac6a34d77a6b3b1fe222393f:

What is the purpose of this `.to_string`?

fuzz/fuzz_targets/encode_decode.rs
apoelstra: I'd also assert that the HRP round-trips.

src/primitives/hrp.rs
```diff
@@ -0,0 +1,420 @@
...
+    /// This does _not_ check that the `hrp` is an in-use HRP within Bitcoin (eg, "bc"), rather it
+    /// checks that the HRP string is valid as per the specification in [BIP-173]:
+    ///
+    /// > The human-readable part, which is intended to convey the type of data, or anything else that
+    /// > is relevant to the reader. This part MUST contain 1 to 83 US-ASCII characters, with each
+    /// > character having a value in the range [33-126]. HRP validity may be further restricted by
+    /// > specific applications.
+    ///
+    /// [BIP-173]: <https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki>
+    pub fn parse(hrp: &str) -> Result<Self, Error> { Self::parse_and_case(hrp).map(|(p, _)| p) }
```
apoelstra, line 45: In 599ce6d089904386ac6a34d77a6b3b1fe222393f:

You can use `.0` as shorthand for `|(p, _)| p`.

src/primitives/hrp.rs
```diff
@@ -0,0 +1,420 @@
...
+        if hrp.len() > MAX_HRP_LEN {
+            return Err(TooLong(hrp.len()));
+        }
+
+        let mut has_lower: bool = false;
+        let mut has_upper: bool = false;
+        for (i, b) in hrp.bytes().enumerate() {
+            // Valid subset of ASCII
+            if !(33..=126).contains(&b) {
+                return Err(InvalidAsciiByte(b));
```
apoelstra, line 65: In 599ce6d089904386ac6a34d77a6b3b1fe222393f:

It would probably be more user-friendly to return a character here instead of a byte.

src/primitives/hrp.rs
```diff
@@ -0,0 +1,420 @@
...
+            }
+            new.buf[i] = b;
+            new.size += 1;
+        }
+
+        let case = match (has_lower, has_upper) {
+            (true, false) => Case::Lower,
+            (false, true) => Case::Upper,
+            (false, false) => Case::None,
+            (_, _) => unreachable!("checked above"),
```
apoelstra, line 85: In 599ce6d089904386ac6a34d77a6b3b1fe222393f:

I think it would be nicer to just move the error return here. It means you lose the short-circuiting but it also means that you eliminate a panic branch.

src/primitives/hrp.rs
```diff
@@ -0,0 +1,420 @@
...
+//! > specific applications.
+//!
+//! ref: [BIP-173](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#user-content-Bech32)
+
+#[cfg(feature = "alloc")]
+use alloc::string::String;
+use core::cmp::Ordering;
+use core::fmt::{self, Write};
+use core::iter::FusedIterator;
+use core::slice::Iter;
```
apoelstra, line 17: In 599ce6d089904386ac6a34d77a6b3b1fe222393f:

I think it's confusing to import this type. I'd import `slice` and refer to it as `slice::Iter` instead.

src/primitives/hrp.rs
```diff
@@ -0,0 +1,420 @@
...
+    fn hash<H: core::hash::Hasher>(&self, h: &mut H) { self.buf.hash(h) }
+}
+
+/// Iterator over bytes (ASCII values) of the human-readable part.
+///
+/// ASCII byte values as they were initially parsed (i.e., in the original case).
+pub struct ByteIter<'b> {
+    iter: Iter<'b, u8>,
+    size: usize, // Total size of the iterator.
+    n: usize,    // Number of characters yielded so far.
```
apoelstra, line 182: In 599ce6d089904386ac6a34d77a6b3b1fe222393f:

Why do we need to keep track of this stuff? Why can't `slice::Iter` do it for us?

fuzz/fuzz_targets/encode_decode.rs
tcharding: ugh, code comments for the win!

fuzz/fuzz_targets/encode_decode.rs
tcharding: > I'd just take the first byte as a "length", try to parse up to that many bytes as an HRP

So that's something I don't understand, even with the tight requirements on the ascii byte range will the fuzzer find data that parses?

fuzz/fuzz_targets/encode_decode.rs
tcharding: derp, what was I even mapping.

src/primitives/hrp.rs
tcharding: double derp, was I on crack yesterday?

src/primitives/hrp.rs
tcharding: I came to the conclusion that converting a byte that was pulled out of a unicode point was not useful for debugging. This happened when I was debugging some of the current bip test code we have and we were using a unicode string that was actually two bytes and then the first byte caused an error and was printed as a char like you suggest, took me ages to work out what was going on.

tcharding: With the raw invalid byte value at least the dev debugging can think straight away "where did that byte come from" instead of wondering where that weird utf-8 char came from and having to convert it to a byte himself - at least that is what I had to do.

src/primitives/hrp.rs
tcharding: How about this so we keep the short-circuiting
```
            (_, _) => return Err(MixedCase), // Defensive programming, this statement is unreachable.
```

src/primitives/hrp.rs
tcharding: Ah yes, that is nicer.

src/primitives/hrp.rs
tcharding: Ha, overthought that didn't I.

src/primitives/hrp.rs
```diff
@@ -0,0 +1,420 @@
...
+pub struct CharIter<'b> {
+    iter: ByteIter<'b>,
+}
+
+impl<'b> Iterator for CharIter<'b> {
+    type Item = char;
+    fn next(&mut self) -> Option<char> {
+        // Usage of unsafe is ok because we uphold the invariant that all
+        // bytes in the HRP are valid ASCII values.
+        self.iter.next().map(|b| unsafe { char::from_u32_unchecked(b as u32) })
```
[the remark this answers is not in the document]
tcharding: Sweeeeet.

fuzz/fuzz_targets/encode_decode.rs
apoelstra: @tcharding yep. It's kinda magical. Basically it will see that whenever it goes out of bounds, this short-circuits the entire logic chain, so it shouldn't do that (or rather, after doing it once, there is nothing more for it to learn). So it is able to very quickly zero in on valid strings for pretty-much any definition of "valid".

src/primitives/hrp.rs
apoelstra: So, I think we should be iterating over `hrp.chars()` in this loop, and then if we get a *char* that's out of bounds, we can stick that in the error. After checking it's in bounds we can safely cast it to a u8.

src/primitives/hrp.rs
apoelstra: Yeah, I'd be cool with that.

src/primitives/hrp.rs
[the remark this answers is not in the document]
apoelstra: Yeah :) I've spent a lot of time grepping the docs for `char` for `Into` and `From` impls for this crate..

src/primitives/hrp.rs
tcharding: Damn, I should trust myself more when reading reviews, we keep track of this because if we use `slice::Iter` it includes all the zeros in the hrp internal buffer.

fuzz/fuzz_targets/encode_decode.rs
tcharding: Woke up this morning thinking about fuzzing, realized I could throw a panic in the fuzz test to see how long it took to hit (had not yet read your post at that stage), it hits it immediately - that is magic, what genius invented that methodology?

tcharding:
Worked out why I couldn't get `duplicate_crash` to work - I didn't realise the output from `./fuzz.sh` over multiple lines was actually a single line if input, I was treating it as multiple lines of input. (I don't remember right now where we were discussing this @apoelstra.)

tcharding:
Changes in force push:
- A bunch of fuzzing fixes including adding a new fuzz target
- Iterate over chars as discussed above

src/primitives/hrp.rs
apoelstra: Then you should create the `slice::Iter` from a slice of the correct size :)

tcharding: oooo, I should trust reviewers more ... :)

tcharding:
Changes in force push:
- Used `iter::Take<slice::Iter<'b, u8>>` for the internal byte iterator so it has the correct length
- Improved the rustdocs to be more uniform
- Simplified the parsing code a little with no change of logic (removed early return since it is an optimisation and makes the code marginally more difficult to read)

fuzz/fuzz_targets/parse_hrp.rs
```diff
@@ -0,0 +1,59 @@
+extern crate bech32;
+
+use bech32::Hrp;
+
+fn do_test(data: &[u8]) {
+    let s = String::from_utf8_lossy(data);
+
+    // Make sure parsing garbage doesn't make us crash (from_utf8_lossy should
+    // contain some garbage, perhaps even invalid chars).
+    let _ = Hrp::parse(&s);
```
apoelstra, line 10: `from_utf8_lossy` will not include any invalid characters, at least not in the UTF8 sense.

I think we should use [str::from_utf8[(https://doc.rust-lang.org/std/str/fn.from_utf8.html) and return on error. This avoids an allocation and avoids mutating the input string, so it should be faster.

fuzz/fuzz_targets/encode_decode.rs
```diff
@@ -13,26 +15,22 @@ fn do_test(data: &[u8]) {
         return;
     }
 
-    let hrp = String::from_utf8_lossy(&data[1..hrp_end])
-        .to_lowercase()
-        .to_string();
+    let hrp = String::from_utf8_lossy(&data[1..hrp_end]).to_lowercase().to_string();
```
apoelstra, line 18: In 179868360b76d5d9a8d806dd5721d51d84e84b48:

Did rustfmt make all these changes to this fuzz unit? They're a little distracting.

I have a bunch of comments about this fuzztest but since it already exists it's probably better to deal with it in a separate PR..

src/primitives/hrp.rs
```diff
@@ -0,0 +1,406 @@
...
+            return Err(TooLong(hrp.len()));
+        }
+
+        let mut has_lower: bool = false;
+        let mut has_upper: bool = false;
+        for (i, c) in hrp.chars().enumerate() {
+            if !c.is_ascii() {
+                return Err(NonAsciiChar(c));
+            }
+            let b: u8 = c.try_into().expect("cannot fail for ASCII values");
```
apoelstra, line 67: In 179868360b76d5d9a8d806dd5721d51d84e84b48:

I think you should just write `let b = c as u8; // cast OK as we just checked that `c` is an ASCII value` which is shorter and eliminates a branch, a panic, and a static string, none of which I really trust the compiler to get rid of.

src/primitives/hrp.rs
```diff
@@ -0,0 +1,406 @@
...
+    fn next(&mut self) -> Option<u8> { self.iter.next().cloned() }
+
+    fn size_hint(&self) -> (usize, Option<usize>) { (self.len(), Some(self.len())) }
+}
+
+impl<'b> ExactSizeIterator for ByteIter<'b> {
+    fn len(&self) -> usize { self.iter.len() }
+}
+
+impl<'b> FusedIterator for ByteIter<'b> {}
```
apoelstra, line 194: In 179868360b76d5d9a8d806dd5721d51d84e84b48:

I guess we ought to implement `DoubleEndedIterator` as well for good measure.

src/primitives/hrp.rs
apoelstra: In 179868360b76d5d9a8d806dd5721d51d84e84b48:

This is fine but I kinda think we should just fold this into `NonAsciiChar` (and rename this to `InvalidChar`). I don't think the user cares whether an invalid character was in the ASCII range or not.

src/primitives/hrp.rs
```diff
@@ -0,0 +1,406 @@
...
+    /// This does _not_ check that the `hrp` is an in-use HRP within Bitcoin (eg, "bc"), rather it
+    /// checks that the HRP string is valid as per the specification in [BIP-173]:
+    ///
+    /// > The human-readable part, which is intended to convey the type of data, or anything else that
+    /// > is relevant to the reader. This part MUST contain 1 to 83 US-ASCII characters, with each
+    /// > character having a value in the range [33-126]. HRP validity may be further restricted by
+    /// > specific applications.
+    ///
+    /// [BIP-173]: <https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki>
+    pub fn parse(hrp: &str) -> Result<Self, Error> { Ok(Self::parse_and_case(hrp)?.0) }
```
apoelstra, line 46: In 179868360b76d5d9a8d806dd5721d51d84e84b48:

Would be nice to add a `parse_unchecked` function that was `const`. It could call `str::as_bytes` and just copy data over. We don't even need to mark it `unsafe` because no single-byte value is an invalid character (or at least, `char::from<u8>()` somehow deals with the ones that are).

Would save people (and us) a lot of `parse("obviouslycorrect").unwrap()` code.

apoelstra: Or if you wanted to preserve invariants we could just check everything is in the 33..126 range and replace any invalid bytes (or the whole string) with `x` or something.

apoelstra:
Done reviewing 179868360b76d5d9a8d806dd5721d51d84e84b48. Looking a lot better!

fuzz/fuzz_targets/parse_hrp.rs
tcharding: oooo, it replaces invalid chars with `?` - that's where those all came from when I tried to iterate over all the zeros in the buffer :)

> Converts a slice of bytes to a string, including invalid characters.

ref: https://doc.rust-lang.org/std/string/struct.String.html#method.from_utf8_lossy

I think your right, I'll try with `from_utf`. Thanks

fuzz/fuzz_targets/encode_decode.rs
tcharding: Sorry about that, I removed the rustfmt changes. PR still has changes to this fuzz test, I took your comments on the other one and applied it to this test. We have to touch this fuzz test in the PR because this PR changes the API.

src/primitives/hrp.rs
tcharding: Sure thing, thanks for the reasoning.

src/primitives/hrp.rs
```diff
@@ -0,0 +1,406 @@
...
+/// Iterator over bytes (ASCII values) of the human-readable part.
+///
+/// ASCII byte values as they were initially parsed (i.e., in the original case).
+pub struct ByteIter<'b> {
+    iter: iter::Take<slice::Iter<'b, u8>>,
+}
+
+impl<'b> Iterator for ByteIter<'b> {
+    type Item = u8;
+    fn next(&mut self) -> Option<u8> { self.iter.next().cloned() }
```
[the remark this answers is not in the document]
tcharding: Done!

src/primitives/hrp.rs
```diff
@@ -0,0 +1,406 @@
...
+
+    /// Returns this human-readable part as a lowercase string.
+    #[cfg(feature = "alloc")]
+    pub fn to_lowercase(&self) -> String { self.lowercase_char_iter().collect() }
+
+    /// Creates a byte iterator over the ASCII byte values (ASCII characters) of this HRP.
+    ///
+    /// If an uppercase HRP was parsed during object construction then this iterator will yield
+    /// uppercase ASCII `char`s. For lowercase bytes see [`Self::lowercase_byte_iter`]
+    pub fn byte_iter(&self) -> ByteIter { ByteIter { iter: self.buf.iter().take(self.size) } }
```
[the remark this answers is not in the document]
tcharding: Used as suggested, thanks. We'll get there in the end ;)

src/primitives/hrp.rs
tcharding: Done for each iterator and tested (noticed I missed on FusedIterator impl also).

src/primitives/hrp.rs
tcharding: I just paniced if the input is invalid.

tcharding:
Thanks for your patience with so many review cycles @apoelstra!

src/primitives/hrp.rs
apoelstra: Interesting, I didn't know you could panic in constfns. I guess you just can't `unwrap`?

apoelstra: I dunno if `unchecked` is the right word now for this :P given that the input is, in fact, checked. But I think it's fine. It signals to the user that they've gotta do their own validation.

apoelstra (approved):
ACK e8a1a5ec6915c87f7cc599a6c47bc000ebf70e3e

apoelstra:
@clarkmoody I'm happy with this PR now. ok to merge?

clarkmoody (approved):
This is really good stuff 👌 

Just a couple wishlist items that don't affect my approval:

- A new function `Hrp::empty()` or `impl Default` that replaces the two `Hrp { buf: [0_u8; MAX_HRP_LEN], size: 0 };` statements.
- (Is this out of scope for this library?) Some constants for Bitcoin HRPs:

```rust
impl Hrp {
    const BITCOIN: Self = Self { buf: [b'b', b'c', 0, 0, 0...], size: 2 };
    // Testnet too
}
```

tcharding:
I added `default` and one usage of it (we cannot call it in the const function).

Can we leave the consts to another PR as we see how the separation between `rust-bech32` and `rust-bitcoin` develops?

apoelstra:
We can leave both to a future PR since they're small enough that they'll be quick to review (and not cause any conflicts).

src/primitives/hrp.rs
apoelstra: So, `panic` is not actually const in 1.48. I found a bug in my local CI setup that was causing the set of compilers to be sometimes collapsed into one. But I don't know why the github CI didn't catch this, and isn't catching this.

`cargo +1.48.0 test` fails right now on master.

PR #104, opened 2023-05-15
Fix build on 1.48.0

apoelstra:
Will update with actual fixes. Just pushing this for now to check that it causes CI to fail.

There are multiple ways this crate is broken on 1.48 right now, but were not caught because of independent bugs in my local CI script and in the Github Actions script.


apoelstra:
cc @tcharding changed the `_unchecked` logic to replace bad characters with `X`s. This might result in mixed-case HRPs but it won't result in bad unicode, so I think this is a safe way to handle bad calls to this function.

contrib/test.sh
```diff
@@ -5,9 +5,6 @@
 set -eu
 set -x
```
tcharding, line 7: There is a whitespace issue with this file, could get fixed in this PR
```diff
diff --git a/contrib/test.sh b/contrib/test.sh
index 76e22ed..698bb7a 100755
--- a/contrib/test.sh
+++ b/contrib/test.sh
@@ -49,8 +49,7 @@ if [ "${DO_FEATURE_MATRIX-false}" = true ]; then
     build_and_test "alloc"
     build_and_test "std alloc"
     # arrayvec breaks the MSRV
-    if [ $MSRV = false
-       ]; then
+    if [ $MSRV = false ]; then
         build_and_test "arrayvec"
         build_and_test "std arrayvec"
         build_and_test "alloc arrayvec"
```

src/primitives/hrp.rs
```diff
@@ -44,25 +44,22 @@ impl Hrp {
     /// [BIP-173]: <https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki>
     pub fn parse(hrp: &str) -> Result<Self, Error> { Ok(Self::parse_and_case(hrp)?.0) }
 
-    /// Parses the human-readable part (see [`Hrp::parse`] for full docs).
-    ///
-    /// # Panics
-    ///
-    /// If the string is invalid as defined by BIP-173.
+    /// Parses the human-readable part (see [`Hrp::parse`] for full docs). Replaces any invalid
+    /// characters with `X`.
```
tcharding, line 48: Since the `_unchecked` was meant to be named after stdlib functions that panic if input is incorrect perhaps we can justify the name with the following:
```
    /// Parses the human-readable part (see [`Hrp::parse`] for full docs).
    ///
    /// Does not check that `hrp` is valid according to BIP-173 but does check for valid ASCII
    /// values, replacing any invalid characters with `X`.
```

apoelstra: Huh. Do you have an example of this? I don't think `_unchecked` functions normally panic. They usually just result in wrong behavior, or in the case of ones marked `unsafe`, cause UB. Panicking sounds like checking to me.

In any case, happy to change the docs to match your suggestion.

tcharding: ooph, should have looked this up before posting, I got two things wrong (what I had in mind is stdlib integers):
- u32 uses `unchecked_foo` not `foo_unchecked`
- they contain UB _not_ panic if input is wrong

> This results in undefined behavior if rhs is larger than or equal to the number of bits in self, i.e. when [checked_shr](https://doc.rust-lang.org/std/primitive.u32.html#method.checked_shr) would return None.

apoelstra: Heh, searching the stdlib docs, it seems that everything *except* the arithmetic opcodes use `_unchecked`, while the arithimetic ones use `uncheckde_`.

But as for UB, I do think that's the most common case. I did consider doing literally no checks, which might've left us in that situation, but it seemed pretty easy to define the behavior in an obviously-wrong way so that we wouldn't need to stick `unsafe` onto the function.

tcharding (approved):
ACK efdc72fee962ee98b7606aa5924d342af3e3337a

apoelstra:
cc @clarkmoody could you ACK this? I guess @tcharding is not a maintainer here.

tcharding:
Poor @tcharding :(

clarkmoody (approved):
ACK from me. If you're using an `_unchecked` function, you should be reading the docs 😄 

The `'X'` is a good alternative to panicking.

PR #105, opened 2023-05-17
Add iter and checksum modules

tcharding:
Add, as yet unused, modules that implement BCH codes and iterators (used for no-alloc access to the bech32 string parts).

Draft because I cannot work out a nice next step. 

I've spent two good chunks of time today and yesterday and I'm stumped how to proceed. Sorry for the Wall Of Text.

1. We cannot use the iterators to implement current traits/types in `lib.rs` because of the use of `AsRef<[u8]>`, which is not compatible with the iterators because we require `ExactSizedIterator`.
2. I'm confused as to what abstraction we are aiming for for this crate
   1. is it a general purpose bech32 crate
   2. is it specifically there to support the `rust-bitcoin` crate i.e., layer one bitcoin
  
  If its (i) then I don't even know what this means, my whole understanding of the crate is based solely on the two bips. If it is (ii) then I don't know how to do a simple-next-step without just re-writing the whole API.

### More points on the abstraction
1. What goes in this crate vs in the `bitcoin::address` module (e.g. who knows about "bc", "tb", "bcrt", valid witness versions, valid witness program length)?
2. Why does `rust-bech32-bitcoin` exist and is the abstraction implied by its existence valid (I know you already said forget about that crate @apoelstra but it effects the abstraction discussion IMO)?

### Moving Forward

1. Can we just clobber the whole API (`lib.rs`) and write an API tailored to bitcoin layer one and let other users just go to `primitives` and use the more clunky but fully functional API?
2. Do we need to just add a `segwit` module that implements the API in (1) and leave `lib.rs` as it is?

If (1) we need buyin from @clarkmoody because its his crate but also because he understands who else is using the crate.
If (2) I have no idea how we ever stabalize the crate because no one in rust-bitcoin will sign off on it.

### DIFF
--- a/src/primitives/checksum.rs
+++ b/src/primitives/checksum.rs
@@ -1,0 +1,202 @@
+// SPDX-License-Identifier: MIT
+
+//! Degree-2 [BCH] code checksum.
+//!
+//! [BCH]: <https://en.wikipedia.org/wiki/BCH_code>
+
+use core::{mem, ops};
+
+use crate::primitives::gf32::Fe32;
+use crate::primitives::hrp::Hrp;
+
+/// Trait defining a particular checksum.
+///
+/// For users, this can be treated as a marker trait; none of the associated data
+/// are end-user relevant.
+pub trait Checksum {
+    /// An unsigned integer type capable of holding a packed version of the generator
+    /// polynomial (without its leading 1) and target residue (which will have the
+    /// same width).
+    ///
+    /// Generally, this is the number of characters in the checksum times 5. So e.g.
+    /// for bech32, which has a 6-character checksum, we need 30 bits, so we can use
+    /// u32 here.
+    ///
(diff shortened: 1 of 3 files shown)


clarkmoody:
The original purpose of the `bech32-bitcoin` crate was to have data structures for Witness Version and do address HRP validation, etc. This library was for the encoding of raw bytes and performing checksums. However, I'm not averse to this library growing to include support for Bitcoin-specific semantics around the structure of our Bech32-encoded strings.

tcharding:
Awesome, cheers - I'll have a go with this strategy.

tcharding:
Closing as we don't need this now, it was only open as draft still for the comment https://github.com/rust-bitcoin/rust-bech32/pull/105#issuecomment-1553036343 but I can remember where that is.

PR #106, opened 2023-05-18
Reduce noise in HTML docs

tcharding:
The "overview" in `primitives::mod` causes the same info to be displayed twice in the HTML rendering of our docs. Just remove it.


clarkmoody (approved):
ACK

apoelstra (approved):
utACK 110ba6c6a314731a182117b51da595c1d3a66e37

PR #110, opened 2023-06-18
Make Bech32Writer constructor public

tcharding:
During comit `10e32e91 Add checksum module` I somehow made the constructor private, this is clearly incorrect. Revert the change by making the constructor public again.


apoelstra:
I think we should just keep moving forward toward deleting `BitWriter` entirely.

PR #111, opened 2023-06-19
Add `iter` module

tcharding:
Add a new `primitives::iter` module that adds various adaptor iterators for encoding with checksums without allocation.

 Use it to implement some of the encode functions and remove the `Bech32Writer`.

Note, this touches the encodes side of the API only, not the decode side. That will follow. Also I deviate from your suggested approach apoelstra, more from tunnel vision while working than for any thought out reason. If you think its better I can remove the "remove `Bech32Writer`" stuff from this and do it last like you suggested.

### DIFF
--- a/embedded/no-allocator/src/main.rs
+++ b/embedded/no-allocator/src/main.rs
@@ -10,7 +10,8 @@
 use panic_halt as _;
 
 use arrayvec::{ArrayString, ArrayVec};
-use bech32::{self, u5, ComboError, FromBase32, ToBase32, Variant, Hrp};
+use bech32::{self, u5, ComboError, Variant, Hrp};
+use bech32::primitives::iter::{Fe32IterExt, ByteIterExt};
 use cortex_m_rt::entry;
 use cortex_m_semihosting::{debug, hprintln};
 
@@ -21,8 +22,9 @@ fn main() -> ! {
     let mut encoded = ArrayString::<30>::new();
 
     let mut base32 = ArrayVec::<u5, 30>::new();
-
-    [0x00u8, 0x01, 0x02].write_base32(&mut base32).unwrap();
+    for fe in [0x00u8, 0x01, 0x02].iter().copied().bytes_to_fes() {
+        base32.push(fe);
+    }
 
     let hrp = Hrp::parse("bech32").unwrap();
 
@@ -38,7 +40,12 @@ fn main() -> ! {
     let (got_hrp, data, variant) =
         bech32::decode_lowercase::<ComboError, _, _>(&encoded, &mut decoded, &mut scratch).unwrap();
     test(got_hrp == hrp);
-    let res = ArrayVec::<u8, 30>::from_base32(&data).unwrap();
+
+    let mut res = ArrayVec::<u8, 30>::new();
+    for byte in data.iter().copied().fes_to_bytes() {
+        res.push(byte);
+    }
+
     test(&res == [0x00, 0x01, 0x02].as_ref());
     test(variant == Variant::Bech32);
 
--- a/embedded/no-allocator/src/main.rs
+++ b/embedded/no-allocator/src/main.rs
@@ -26,9 +26,7 @@ fn main() -> ! {
 
     let hrp = Hrp::parse("bech32").unwrap();
 
-    bech32::encode_to_fmt(&mut encoded, hrp, &base32, Variant::Bech32)
-        .unwrap()
-        .unwrap();
+    bech32::encode_to_fmt(&mut encoded, hrp, &base32, Variant::Bech32).unwrap();
     test(&*encoded == "bech321qqqsyrhqy2a");
 
     hprintln!("{}", encoded).unwrap();
--- a/embedded/no-allocator/src/main.rs
+++ b/embedded/no-allocator/src/main.rs
@@ -21,14 +21,11 @@ use cortex_m_semihosting::{debug, hprintln};
 fn main() -> ! {
     let mut encoded = ArrayString::<30>::new();
 
-    let mut base32 = ArrayVec::<u5, 30>::new();
-    for fe in [0x00u8, 0x01, 0x02].iter().copied().bytes_to_fes() {
-        base32.push(fe);
-    }
+    let base32 = [0x00u8, 0x01, 0x02].iter().copied().bytes_to_fes().collect::<ArrayVec<u5, 30>>();
 
     let hrp = Hrp::parse("bech32").unwrap();
 
-    bech32::encode_to_fmt(&mut encoded, hrp, &base32, Variant::Bech32).unwrap();
+    bech32::encode_to_fmt(&mut encoded, &hrp, base32, Variant::Bech32).unwrap();
     test(&*encoded == "bech321qqqsyrhqy2a");
 
     hprintln!("{}", encoded).unwrap();
@@ -41,10 +38,7 @@ fn main() -> ! {
         bech32::decode_lowercase::<ComboError, _, _>(&encoded, &mut decoded, &mut scratch).unwrap();
     test(got_hrp == hrp);
 
-    let mut res = ArrayVec::<u8, 30>::new();
-    for byte in data.iter().copied().fes_to_bytes() {
-        res.push(byte);
-    }
+    let res = data.iter().copied().fes_to_bytes().collect::<ArrayVec<u8, 30>>();
 
     test(&res == [0x00, 0x01, 0x02].as_ref());
     test(variant == Variant::Bech32);
--- a/embedded/no-allocator/src/main.rs
+++ b/embedded/no-allocator/src/main.rs
@@ -26,7 +26,7 @@ fn main() -> ! {
 
     let hrp = Hrp::parse("bech32").unwrap();
 
-    bech32::encode_to_fmt_anycase(&mut encoded, hrp, &base32, Variant::Bech32)
+    bech32::encode_to_fmt(&mut encoded, hrp, &base32, Variant::Bech32)
         .unwrap()
         .unwrap();
     test(&*encoded == "bech321qqqsyrhqy2a");
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -438,44 +438,6 @@ pub fn encode_to_fmt<T: AsRef<[u5]>>(
     Ok(Ok(()))
 }
 
-/// Encode a bech32 payload to an [fmt::Write], but with any case.
-/// This method is intended for implementing traits from [core::fmt] without [std].
-///
-/// See `encode_to_fmt` for meaning of errors.
-pub fn encode_to_fmt_anycase<T: AsRef<[u5]>>(
-    fmt: &mut dyn fmt::Write,
-    hrp: Hrp,
-    data: T,
-    variant: Variant,
-) -> Result<fmt::Result, Error> {
-    match variant {
-        Variant::Bech32 => {
-            let res = Bech32Writer::<Bech32>::new(hrp, fmt);
-            match res {
-                Ok(mut writer) => {
-                    Ok(writer.write(data.as_ref()).and_then(|_| {
-                        // Finalize manually to avoid panic on drop if write fails
-                        writer.finalize()
-                    }))
-                }
-                Err(e) => Ok(Err(e)),
-            }
-        }
-        Variant::Bech32m => {
-            let res = Bech32Writer::<Bech32m>::new(hrp, fmt);
-            match res {
-                Ok(mut writer) => {
-                    Ok(writer.write(data.as_ref()).and_then(|_| {
-                        // Finalize manually to avoid panic on drop if write fails
-                        writer.finalize()
-                    }))
-                }
-                Err(e) => Ok(Err(e)),
-            }
-        }
-    }
-}
-
 /// Encodes a bech32 payload without a checksum to a writer ([`fmt::Write`]).
 ///
 /// This method is intended for implementing traits from [`std::fmt`].
@@ -1221,7 +1183,7 @@ mod tests {
         [0x00u8, 0x01, 0x02].write_base32(&mut base32).unwrap();
 
         let bech32_hrp = Hrp::parse("bech32").expect("bech32 is valid");
-        encode_to_fmt_anycase(&mut encoded, bech32_hrp, &base32, Variant::Bech32).unwrap().unwrap();
+        encode_to_fmt(&mut encoded, bech32_hrp, &base32, Variant::Bech32).unwrap().unwrap();
         assert_eq!(&*encoded, "bech321qqqsyrhqy2a");
 
         println!("{}", encoded);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -114,7 +114,7 @@ impl<'a, Ck: Checksum> Bech32Writer<'a, Ck> {
     ///
     /// This is a rather low-level API and doesn't check the HRP or data length for standard
     /// compliance.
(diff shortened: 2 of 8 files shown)


tcharding:
CC @Ericson2314 because he added the `arrayvec` stuff. There has been some heavy re-working (still ongoing) of this crate mate, just cc'ing you out of courtesy.

tcharding:
In case anyone reads all the force pushes, I originally removed the `arrayvec` stuff too but after battling with CI (because of fuzzing and alloc/no-alloc tests) I realize we need to do decodeing before removing that stuff.

src/primitives/iter.rs
```diff
@@ -0,0 +1,501 @@
...
+}
+
+/// Returns the number of Fe32's encoded by `n` bytes.
+fn bytes_len_to_fes_len(n: usize) -> usize {
+    let bits = n * 8;
+    if bits % 5 != 0 {
+        bits / 5 + 1 // +1 because we pad the fes.
+    } else {
+        bits / 5
+    }
```
apoelstra, line 348: In 0c6613106fa624db037598dab5e9a19634acd2d0:

It would be clearer and more idiomatic to just write `(bits + 4) / 5` if you mean "bits / 5, rounded up". TBH I think this function could be replaced by inline code and doesn't need any unit tests.

src/primitives/iter.rs
```diff
@@ -0,0 +1,501 @@
...
+impl<I, Ck> ChecksumIter<I, Ck>
+where
+    I: Iterator<Item = Fe32>,
+    Ck: Checksum,
+{
+    /// Helper function to input an HRP into the underlying checksum engine of the iterator.
+    ///
+    /// This function is infallible, but if you feed it a non-ASCII `hrp` it probably
+    /// will cause your checksum engine to produce useless results.
+    pub fn with_checksummed_hrp(mut self, hrp: &Hrp) -> Self {
```
apoelstra, line 254: In 0c6613106fa624db037598dab5e9a19634acd2d0:

Maybe in the doccomment also mention that "if you use this after the iterator has already yielded elements" that this probably will also produce useless results.

src/primitives/iter.rs
```diff
@@ -0,0 +1,501 @@
...
+///
+/// (If the total number of bits is not a multiple of 8, any trailing bits are dropped.)
+fn fes_len_to_bytes_len(n: usize) -> usize { n * 5 / 8 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// Asserts two iterators are equal.
+    fn check_iter_eq<I, J, T>(mut i: I, mut j: J)
```
apoelstra, line 361: In 0c6613106fa624db037598dab5e9a19634acd2d0:

Could this whole function be replaced with [Iterator::eq](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.eq)?

src/lib.rs
```diff
@@ -497,10 +492,27 @@ impl Variant {
 /// * Deviations from standard.
 /// * No length limits are enforced for the data part.
 #[cfg(feature = "alloc")]
```
apoelstra, line 429: In 1facf64fcd706c3c2a4f6c21349efcdcd308cd93:

You can remove the `cfg(alloc)` from this function now. You may also want to reformat the doccomment to have `Deviations from standard` be a heading, like you did on the other one.

src/primitives/iter.rs
tcharding: Done, same for the other direction (fes to bytes)

src/primitives/iter.rs
tcharding: Done!

src/primitives/iter.rs
tcharding: yep, seems so. Done.

src/lib.rs
tcharding: This is on `encode`, right? If so it returns a `String` (I checked all feature gating and it looks correct but I'm brain fried as f*** right now :)

tcharding: I fixed the docs.

apoelstra: > This is on `encode`, right? If so it returns a `String` (I checked all feature gating and it looks correct.)

Derp, you're right.

apoelstra (approved):
ACK b29c0a407bae4b1c639a0b877783c3c39ceeb1bc

apoelstra:
@clarkmoody wanna take a look at this?

tcharding:
Hey @apoelstra I messed around with the `size_hint` functions from what you originally wrote, can you take a look at them at some stage just to make sure you are happy. FTR I learned that returning `(0, None)` is correct for all iterators and that the return values are not meant to be relied upon anyway so I think its not super important. (We were returning a value too high for `min` in one function which is what got me started on investigating them all.)

apoelstra:
@tcharding ok, looking more closely

* For `ByteToFeIter` I think the +1 should be inside the call to `bytes_len_to_fes_len` not outside it
* `FeToByteIter` looks right
* `WitnessVersionIter` I think should add 1 to `hint.0`.
* `ChecksumIter` looks right
* `HrpCharIter` looks right

tcharding:
Cool, thanks. I pulled `bytes_len_to_fes_len` back out into a function because the closure looked too messy. Fixed witness version iter as suggested.

tcharding:
I added two additional patches because removing `ToBase32` and `FromBase32` displays the usefulness of the `iter` module.

src/primitives/iter.rs
```diff
@@ -0,0 +1,414 @@
...
+            checksum_remaining: Ck::CHECKSUM_LENGTH,
+            checksum_engine: checksum::Engine::new(),
+        }
+    }
+
+    /// Adapts the Fe32 iterator to output characters using `hrp` for the human-readable part.
+    ///
+    /// Note, `hrp` is expected to be the same as that fed into the checksum engine with
+    /// `with_checksummed_hrp`.
+    fn hrp_char(self, hrp: &Hrp) -> HrpCharIter<'_, Self> {
```
clarkmoody, line 152: Is there any reason to keep `with_checksummed_hrp` and `hrp_char` as separate functions? Why not combine into a single adapter? Are there reasonable use cases where these two functions are not always called together?

src/lib.rs
```diff
@@ -800,6 +679,7 @@ mod tests {
     use arrayvec::ArrayString;
 
     use super::*;
+    use crate::primitives::iter::{ByteIterExt, Fe32IterExt};
```
apoelstra, line 682: In e03f93231e66c703c44bdbfd1218f6d19029a697:

I believe `Fe32IterExt` is not actually used.

embedded/no-allocator/src/main.rs
```diff
@@ -38,7 +40,12 @@ fn main() -> ! {
     let (got_hrp, data, variant) =
         bech32::decode_lowercase::<ComboError, _, _>(&encoded, &mut decoded, &mut scratch).unwrap();
     test(got_hrp == hrp);
-    let res = ArrayVec::<u8, 30>::from_base32(&data).unwrap();
+
+    let mut res = ArrayVec::<u8, 30>::new();
+    for byte in data.iter().copied().fes_to_bytes() {
+        res.push(byte);
+    }
```
apoelstra, line 47: In e03f93231e66c703c44bdbfd1218f6d19029a697:

Every one of these `new()` followed by `push` in a loop tests can be replaced with `.collect()` on the iterator.

src/primitives/iter.rs
apoelstra: As an iterator adaptor, `hrp_char` doesn't know that there's a checksum engine underneath that it could be feeding data into. So it just outputs the data but doesn't do anything about the checksum.

Meanwhile `with_checksummed_hrp` is a method specifically on `ChecksumIter`. So it can feed data into the checksum engine, but it doesn't output it, presumably because `hrp_char` will do it for you.

apoelstra: Ah, the other distinction is that `HrpCharIter` yields characters while `ChecksumIter` yields `Fe32`s. This is probably why we have separate functions.

apoelstra: My vote is:

* Drop `hrp_char` from `Fe32IterExt`. This is a confusing method to have on a general Fe32 iterator
* Change `with_checksummed_hrp` to yield a new iterator type `HrpChecksumIter` which yields characters rather than Fe32s, yields the HRP as character data, and simultaneously feeds it into the underlying checksum engine
* Rename `with_checksummed_hrp` to just `with_hrp`

Good call complaining about this :)

embedded/no-allocator/src/main.rs
tcharding: Interesting, I was mentally associating `collect` with "allocate a new thing" - I didn't think about exactly what it desugars to, I'll give your suggestion a go and see if it can handle `ArrayVec`.

src/lib.rs
tcharding: Clippy should warn us then, I'll investigate.

src/primitives/iter.rs
tcharding: This is an interesting question, the fact that the last 3 calls in this combinator chain are all segwit specific has been bothering me. Also I don't see any use case where one leaves out any of them, basically one either iterates over the fes returned by `bytes_to_fes` (can map to chars) or else one adds the hrp, checksum etc.
```rust
//! let iterator = witness_prog
//!     .iter()
//!     .copied() // Iterate over bytes.
//!     .bytes_to_fes() // Convert bytes to field elements in-line.
//!     .with_witness_v0() 
//!     .checksum::<Bech32>() // Convert to a [`ChecksumIter`] (append a bech32 checksum).
//!     .with_checksummed_hrp(&hrp) // Feed HRP into the checksum.
//!     .hrp_char(&hrp); // Turn the fe stream into a char stream with HRP.
//! let hrpstring: String = iterator.collect();
```

We could have a different design where one adapts an `Fe32` iterator straight into an iterator that iterates over chars, and includes everything (hrp, checksum etc.). All the bip test vectors and `lib.rs` use cases are possible if one chains the witness version onto the front of the fe32 iterator
```rust
let iterator = iter::once(Fe32::Q)
    .chain(
        witness_prog
            .iter()
            .copied()
            .bytes_to_fes()
    )
    .char_iter::<Bech32>(hrp)
```

Note that once we add the final `segwit` API no-one should ever have to write the `iter::once` chain themselves.

embedded/no-allocator/src/main.rs
apoelstra: RIght -- it's a function which wraps `FromIterator::from_iter` which can be implemented on all sorts of collections, whether or not they allocate. Though it does look like all the stdlib implementors are allocating except for the one that lets you collect `()`s into a single `()`.

src/primitives/iter.rs
tcharding: hmmm, that `chain` is ugly, it may be better to have the witness version as a parameter on `char_iter`.

apoelstra: `.checksum::<Bech32>()` is not segwit specific. It takes a stream of field elements and suffixes a given checksum. It's one of the most general things that this library does.

I think it's reasonable to combine `.with_checksummed_hrp` and `.hrp_char` because these methods really don't make sense without each other (sometimes you do want an already-constructed checksum engine, but this is fairly rare, and for this we should just provide a separate `.checksum_with_engine()` adaptor).

But `.checksum()` is really important whenever you're working with field elements and don't want to convert them to characters. i.e. literally any use of field elements except for display. Including verifying checksums.

> We could have a different design where one adapts an `Fe32` iterator straight into an iterator that iterates over chars, and includes everything (hrp, checksum etc.). All the bip test vectors and `lib.rs` use cases are possible if one chains the witness version onto the front of the fe32 iterator

I really strongly dislike this and I'd fork the library if that's what we would up doing.

> All the bip test vectors and `lib.rs` use cases are possible if one chains the witness version onto the front of the fe32 iterator

Yes, but `.chain` absolutely sucks, as you can see, which is why we have `WitnessVersionIterator`.

src/lib.rs
apoelstra: There is a warning.

src/primitives/iter.rs
tcharding: Oh shiiiit, this idea turned out super nice - pushing as a separate patch you guys to look at. Will squash if it gets a concept ack.

embedded/no-allocator/src/main.rs
tcharding: It works, its in this PR now. FTR it panics if the iterator is too long but its only in test code.

> Panics if the number of elements in the iterator exceeds the arrayvec’s capacity.

https://doc.rust-lang.org/nightly/core/iter/trait.FromIterator.html#tymethod.from_iter

src/primitives/iter.rs
apoelstra: If I have a vector of `Fe32`s with a checksum at the end, how can I verify that checksum using this API?

tcharding: One could write something like this (do you want me to add this, I didn't test it for correctness):
```rust

fn validate_checksum<Ck: Checksum>(v: Vec<Fe32>) -> bool {
    if v.len() < Ck::CHECKSUM_LENGTH {
        panic!("to short");
    }
    let chk_boundry = v.len() - Ck::CHECKSUM_LENGTH;
    let data = &v[..chk_boundry];
    let chk = &v[chk_boundry..];

    let mut engine = checksum::Engine::<Ck>::new();
    for fe in data.iter() {
        engine.input_fe(*fe);
    }
    engine.input_target_residue();

    let mut checksum_remaining = Ck::CHECKSUM_LENGTH;
    for c in chk.iter() {
        if *c != Fe32(engine.residue().unpack(checksum_remaining)) {
            return false
        }
        checksum_remaining -= 1;
    }
    true
}
```

tcharding: Although note that currently validating check sums is no where in the code because its in `hrpstring`.

tcharding: So I wrote up some code how to do it then finally realized you mean that by removing the `ChecksumIter` I'm making this hard. Perhaps we should keep the `iter` module as you originally had it and add the new `CharIter` on top of it so that users have both options? But that does lead us back to @clarkmoody's question, "is there a usecase for only `with_checksummed_hrp` and not `hrp_char`"? From an engineering perspective, its not trivial to combine these two adaptors (at least I don't see an easy way).

apoelstra: `with_checksummed_hrp` is useful for verifying checksums. Then you don't want to convert to char, you want to run all your data through it and check the residue of the underlying engine.

Let's switch to Signal for a bit and see if we can hash something out.

src/lib.rs
```diff
@@ -395,66 +193,34 @@ enum Case {
...
-                }
-                Err(e) => Ok(Err(e)),
-            }
-        }
+        Variant::Bech32 =>
+            for c in data.as_ref().iter().copied().char_iter::<Bech32>(hrp, None) {
+                fmt.write_char(c)?;
+            },
+        Variant::Bech32m =>
+            for c in data.as_ref().iter().copied().char_iter::<Bech32>(hrp, None) {
```
clarkmoody, line 211: Should this be `char_iter::<Bech32m>` here?

tcharding: Yes, I"ll add a unit test also, this shouldn't have got through CI. Thanks man.

tcharding:
Superseded by #113

src/primitives/iter.rs
apoelstra: For the record -- @tcharding and I had a phone call and agreed that his approach is actually correct here. Separately, we discussed #112 and the `data_iter` method there, and realized that we needed to merge the two PRs because all these "what are the proper iterator APIs" questions need to be answered at once if Clark is to have a fair chance at reviewing it.

PR #113, opened 2023-06-27
Add `iter` and `hrpstring`

tcharding:
(Just the last patch, the rest is in #116)

Currently there are a bunch of types and traits in `lib.rs` for converting to/from base32 (incl. doing so without allocating).

Observe that instead of custom types we can use iterator adapters to achieve this. Some benefits are:

- More composabality
- More "rusty"
- Uses the new `checksum` module (i.e, better separation of concerns)
- Hopefully more maintainable

Add two new sub-modules to the `primitives` module:

- `iter`: Implement various adaptor iterators for allocationless checksumming etc.
- `hrpstring`: Add the `Parsed` struct that can be used for parsing bech32 address strings.

Remove everything that is now not needed from `lib.rs` while making an effort not to change the API too much i.e., keep the current public functions more or less as they are, with the following notable exceptions:

- `encode_to_fmt` no longer returns a result
- `encode` now returns a `fmt::Error` as the `Result` error variant (instead of `Error`).
- `decode` returns a `Parsed` instead of an HRP and vector of u5s, to get the HRP/u5's one can use the `hrp()` and `fe32_iter()` functions on `Parsed`.
- `decode_without_checksum` return value changed as per `decode`.
- The crate level `Error` type is totally re-written.

All the recently added but as-yet-unreleased `arrayvec` stuff which was added to improve alloc free encoding/decoding is removed.

## Complete list of the public types/functions removed in this patch

(This functionality is now provided by `primitives`.)

- traits:
  - `WriteBase32`
  - `WriteBase256` (unreleased)
  - `FromBase32`
  - `ToBase32`
  - `Base32Len`
  - `CheckBase32`

- types (and their trait implementations):
  - `Bech32Writer`
  - `ComboError` (unreleased)
  - `Case` (exist still but is private)

- functions:
  - `convert_bits`
  - `convert_bits_in` (unreleased)
  - `encode_to_fmt_anycase` (unreleased)
  - `decode_lowercase` (unreleased)

### DIFF
--- a/fuzz/fuzz_targets/encode_decode.rs
+++ b/fuzz/fuzz_targets/encode_decode.rs
@@ -3,7 +3,7 @@ extern crate bech32;
 use std::convert::TryFrom;
 use std::str;
 
-use bech32::Hrp;
+use bech32::{Fe32, Hrp};
 
 fn do_test(data: &[u8]) {
     if data.len() < 1 {
@@ -33,12 +33,13 @@ fn do_test(data: &[u8]) {
             match Hrp::parse(&s) {
                 Err(_) => return,
                 Ok(hrp) => {
-                    if let Ok(data_str) = bech32::encode(hrp, &dp, variant).map(|b32| b32.to_string()) {
-                        let decoded = bech32::decode(&data_str);
-                        let b32 = decoded.expect("should be able to decode own encoding");
+                    let encoded = bech32::encode(hrp, &dp, variant).expect("failed to encode");
+                    let (parsed, variant) = bech32::decode(&encoded).expect("failed to decode");
 
-                        assert_eq!(bech32::encode(b32.0, &b32.1, b32.2).unwrap(), data_str);
-                    }
+                    let hrp = parsed.hrp();
+                    let data = parsed.fe32_iter().collect::<Vec<Fe32>>();
+
+                    assert_eq!(bech32::encode(hrp, data, variant).expect("failed to encode"), encoded);
                 }
             }
         }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -434,9 +434,8 @@ pub fn encode_to_fmt_anycase<T: AsRef<[u5]>>(
 ///
 /// This method is intended for implementing traits from [`std::fmt`].
 ///
-/// # Errors
+/// # Deviations from standard.
 ///
-/// * Deviations from standard.
 /// * No length limits are enforced for the data part.
 pub fn encode_without_checksum_to_fmt<T: AsRef<[u5]>>(
     fmt: &mut dyn fmt::Write,
@@ -484,9 +483,8 @@ impl Variant {
 
 /// Encodes a bech32 payload to string.
 ///
-/// # Errors
+/// # Deviations from standard.
 ///
-/// * Deviations from standard.
 /// * No length limits are enforced for the data part.
 #[cfg(feature = "alloc")]
 pub fn encode<T: AsRef<[u5]>>(hrp: Hrp, data: T, variant: Variant) -> Result<String, Error> {
@@ -497,9 +495,8 @@ pub fn encode<T: AsRef<[u5]>>(hrp: Hrp, data: T, variant: Variant) -> Result<Str
 
 /// Encodes a bech32 payload to string without the checksum.
 ///
-/// # Errors
+/// # Deviations from standard.
 ///
-/// * Deviations from standard.
 /// * No length limits are enforced for the data part.
 #[cfg(feature = "alloc")]
 pub fn encode_without_checksum<T: AsRef<[u5]>>(hrp: Hrp, data: T) -> Result<String, Error> {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -11,24 +11,6 @@
 //!
 //! The original description in [BIP-0173](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki)
 //! has more details. See also [BIP-0350](https://github.com/bitcoin/bips/blob/master/bip-0350.mediawiki).
-//!
-#![cfg_attr(
-    feature = "std",
-    doc = "
-# Examples
-```
-use bech32::{self, FromBase32, ToBase32, Variant, Hrp};
-let hrp = Hrp::parse(\"bech32\").expect(\"bech32 is valid\");
-let encoded = bech32::encode(hrp, vec![0x00, 0x01, 0x02].to_base32(), Variant::Bech32).unwrap();
-assert_eq!(encoded, \"bech321qqqsyrhqy2a\".to_string());
-let (hrp, data, variant) = bech32::decode(&encoded).unwrap();
-assert_eq!(hrp.to_string(), \"bech32\");
-assert_eq!(Vec::<u8>::from_base32(&data).unwrap(), vec![0x00, 0x01, 0x02]);
-assert_eq!(variant, Variant::Bech32);
-```
-"
-)]
-//!
 
 #![cfg_attr(all(not(feature = "std"), not(test)), no_std)]
 // Experimental features we need.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -30,404 +30,46 @@ extern crate core;
 
 #[cfg(all(feature = "alloc", not(feature = "std"), not(test)))]
 use alloc::{string::String, vec::Vec};
-use core::convert::{Infallible, TryFrom};
-use core::{fmt, mem};
+use core::fmt;
 
-pub use crate::primitives::checksum::Checksum;
-use crate::primitives::checksum::{self, PackedFe32};
-use crate::primitives::hrp;
+// Re-export all types from `primitives` modules that appear in the public API.
+pub use crate::primitives::gf32::Fe32;
 pub use crate::primitives::hrp::Hrp;
-pub use crate::primitives::{Bech32, Bech32m};
+pub use crate::primitives::hrpstring::Parsed;
+// Also re-export the iter extensions because the conversion methods are core functionality.
+pub use crate::primitives::iter::{ByteIterExt, Fe32IterExt};
+pub use crate::primitives::{Bech32, Bech32m, NoChecksum};
 
 mod error;
 pub mod primitives;
 
-#[cfg(feature = "arrayvec")]
-use arrayvec::{ArrayVec, CapacityError};
 pub use primitives::gf32::Fe32 as u5;
-
-/// Interface to write `u5`s into a sink.
-pub trait WriteBase32 {
-    /// Write error.
-    type Error: fmt::Debug;
-
-    /// Writes a `u5` slice to `self`.
-    fn write(&mut self, data: &[u5]) -> Result<(), Self::Error> {
-        for b in data {
-            self.write_u5(*b)?;
-        }
-        Ok(())
-    }
-
-    /// Writes a single `u5`.
-    fn write_u5(&mut self, data: u5) -> Result<(), Self::Error> { self.write(&[data]) }
-}
-
-/// Interface to write `u8`s into a sink
-///
-/// Like `std::io::Writer`, but because the associated type is no_std compatible.
-pub trait WriteBase256 {
-    /// Write error.
-    type Error: fmt::Debug;
-
-    /// Writes a `u8` slice.
-    fn write(&mut self, data: &[u8]) -> Result<(), Self::Error> {
-        for b in data {
-            self.write_u8(*b)?;
-        }
-        Ok(())
-    }
-
-    /// Writes a single `u8`.
-    fn write_u8(&mut self, data: u8) -> Result<(), Self::Error> { self.write(&[data]) }
-}
-
-const CHECKSUM_LENGTH: usize = 6;
-
-/// Allocationless Bech32 writer that accumulates the checksum data internally and writes them out
-/// in the end.
-pub struct Bech32Writer<'a, Ck: Checksum> {
-    formatter: &'a mut dyn fmt::Write,
-    engine: checksum::Engine<Ck>,
-}
-
-impl<'a, Ck: Checksum> Bech32Writer<'a, Ck> {
-    /// Creates a new writer that can write a bech32 string without allocating itself.
-    ///
-    /// This is a rather low-level API and doesn't check the HRP or data length for standard
-    /// compliance.
-    fn new(hrp: Hrp, fmt: &'a mut dyn fmt::Write) -> Result<Bech32Writer<'a, Ck>, fmt::Error> {
-        let mut engine = checksum::Engine::new();
-        engine.input_hrp(&hrp);
-
-        for c in hrp.lowercase_char_iter() {
-            fmt.write_char(c)?;
-        }
-        fmt.write_char(SEP)?;
-
-        Ok(Bech32Writer { formatter: fmt, engine })
-    }
-
-    /// Writes out the checksum at the end.
-    ///
-    /// If this method isn't explicitly called this will happen on drop.
-    pub fn finalize(mut self) -> fmt::Result {
-        self.write_checksum()?;
-        mem::forget(self);
-        Ok(())
-    }
-
-    /// Calculates and writes a checksum to `self`.
-    fn write_checksum(&mut self) -> fmt::Result {
-        self.engine.input_target_residue();
-
-        let mut checksum_remaining = self::CHECKSUM_LENGTH;
-        while checksum_remaining > 0 {
-            checksum_remaining -= 1;
-
-            let fe = u5::try_from(self.engine.residue().unpack(checksum_remaining))
-                .expect("unpack returns valid field element");
-            self.formatter.write_char(fe.to_char())?;
-        }
-
-        Ok(())
-    }
-}
-
-impl<'a, Ck: Checksum> WriteBase32 for Bech32Writer<'a, Ck> {
-    type Error = fmt::Error;
-
-    fn write_u5(&mut self, data: u5) -> fmt::Result {
-        self.engine.input_fe(data);
-        self.formatter.write_char(data.to_char())
-    }
-}
-
-impl<'a, Ck: Checksum> Drop for Bech32Writer<'a, Ck> {
-    fn drop(&mut self) {
-        self.write_checksum().expect("Unhandled error writing the checksum on drop.")
-    }
-}
-
-/// Parses/converts base32 slice to `Self`.
-///
-/// This trait is the reciprocal of `ToBase32`.
-pub trait FromBase32: Sized {
-    /// The associated error which can be returned from parsing (e.g. because of bad padding).
-    type Error;
-
-    /// Converts a base32 slice to `Self`.
-    fn from_base32(b32: &[u5]) -> Result<Self, Self::Error>;
-}
-
-macro_rules! write_base_n {
-    { $tr:ident, $ty:ident, $meth:ident } => {
-        #[cfg(feature = "arrayvec")]
-        impl<const L: usize> $tr for ArrayVec<$ty, L> {
-            type Error = CapacityError;
-
-            fn write(&mut self, data: &[$ty]) -> Result<(), Self::Error> {
-                self.try_extend_from_slice(data)?;
-                Ok(())
-            }
-
-            fn $meth(&mut self, data: $ty) -> Result<(), Self::Error> {
-                self.push(data);
-                Ok(())
-            }
-        }
-
-        #[cfg(feature = "alloc")]
-        impl $tr for Vec<$ty> {
-            type Error = Infallible;
-
-            fn write(&mut self, data: &[$ty]) -> Result<(), Self::Error> {
-                self.extend_from_slice(data);
-                Ok(())
-            }
-
-            fn $meth(&mut self, data: $ty) -> Result<(), Self::Error> {
-                self.push(data);
-                Ok(())
-            }
-        }
-    }
-}
-
-write_base_n! { WriteBase32, u5, write_u5 }
-write_base_n! { WriteBase256, u8, write_u8 }
-
-#[cfg(feature = "arrayvec")]
-#[derive(Clone, Debug, PartialEq, Eq)]
-/// Combination of Errors for use with array vec
-pub enum ComboError {
-    /// Error from this crate
-    Bech32Error(Error),
-    /// Error from `arrayvec`.
-    WriteError(CapacityError),
-}
-#[cfg(feature = "arrayvec")]
-impl From<Error> for ComboError {
-    fn from(e: Error) -> ComboError { ComboError::Bech32Error(e) }
-}
-#[cfg(feature = "arrayvec")]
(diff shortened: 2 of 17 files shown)


fuzz/fuzz_targets/encode_decode.rs
```diff
@@ -33,12 +33,17 @@ fn do_test(data: &[u8]) {
...
                 Err(_) => return,
                 Ok(hrp) => {
-                    if let Ok(data_str) = bech32::encode(hrp, &dp, variant).map(|b32| b32.to_string()) {
-                        let decoded = bech32::decode(&data_str);
-                        let b32 = decoded.expect("should be able to decode own encoding");
+                    let encoded = bech32::encode(hrp, &dp, variant).expect("failed to encode");
+                    let decoded = bech32::decode(&encoded);
+                    let (parsed, variant) = match decoded {
+                        Ok((parsed, variant)) => (parsed, variant),
+                        Err(_) => return,
```
apoelstra, line 40: In e2527dc19c3e8e9a4cd678a955fc5b161e175c87:

Decoding should never fail, if this came from the encoding of something we just parsed, right?

fuzz/fuzz_targets/encode_decode.rs
```diff
@@ -33,12 +33,17 @@ fn do_test(data: &[u8]) {
             match Hrp::parse(&s) {
                 Err(_) => return,
                 Ok(hrp) => {
-                    if let Ok(data_str) = bech32::encode(hrp, &dp, variant).map(|b32| b32.to_string()) {
-                        let decoded = bech32::decode(&data_str);
-                        let b32 = decoded.expect("should be able to decode own encoding");
+                    let encoded = bech32::encode(hrp, &dp, variant).expect("failed to encode");
```
apoelstra, line 36: In e2527dc19c3e8e9a4cd678a955fc5b161e175c87:

Should we assert that the encoding here matches the original string? At least, up to the length of the encoded data?

src/lib.rs
```diff
@@ -82,101 +82,8 @@ pub trait WriteBase256 {
...
-/// This trait is the reciprocal of `ToBase32`.
-pub trait FromBase32: Sized {
-    /// The associated error which can be returned from parsing (e.g. because of bad padding).
-    type Error;
-
-    /// Converts a base32 slice to `Self`.
-    fn from_base32(b32: &[u5]) -> Result<Self, Self::Error>;
-}
-
 macro_rules! write_base_n {
```
apoelstra, line 85: In e2527dc19c3e8e9a4cd678a955fc5b161e175c87:

Might as well delete this macro, and the `WriteBase*` traits, so we can have a solid sea of red here. You've removed every usage of these traits and arleady crossed the "breaking API change" rubicon.

apoelstra: This requires you delete the `convert_bits` function and all its unit tests. I'll leave it up to you if you want to do that here. It's not necessary for this PR to compile so maybe you want to keep the diff minimal.

Ditto the `CheckBase32` trait and its tests.

src/lib.rs
```diff
@@ -505,180 +248,26 @@ pub fn encode_without_checksum<T: AsRef<[u5]>>(hrp: Hrp, data: T) -> Result<Stri
...
-        Some(variant) => {
-            // Remove checksum from data payload
-            data.truncate(data.len() - CHECKSUM_LENGTH);
-
-            Ok((hrp_lower, data, variant))
-        }
-        None => Err(Error::InvalidChecksum),
+    match Parsed::new::<Bech32>(s) {
+        Ok(p) => Ok((p, Variant::Bech32)),
+        Err(e) => Err(Error::Hrpstring(e)),
```
apoelstra, line 258: In e2527dc19c3e8e9a4cd678a955fc5b161e175c87:

Is there any way with this API to avoid verifying the entire checksum twice?

src/primitives/hrpstring.rs
```diff
@@ -0,0 +1,447 @@
...
+use crate::write_err;
+
+/// Separator between the hrp and payload (as defined by BIP-173).
+const SEP: char = '1';
+
+/// An HRP string that has been parsed and had the checksum validated.
+///
+/// Pre-parsing an HRP string only checks validity of the characters, it does not validate the checksum in any way - to validate convert to [`Parsed`].
+#[derive(Debug)]
+pub struct Parsed<'s> {
```
apoelstra, line 31: In e2527dc19c3e8e9a4cd678a955fc5b161e175c87:

Here is a proposal:

* We make `Parsed` generic over the `Ck` by adding a phantomdata. This makes the API slightly less flexible but I think is worth it in clarity.
* We add an internal `checksum` field which stores the ASCII values of the checksum.
* Then `new` and `new_with_witness_version` would take `Ck` implicitly from the type that is being constructed.
* We make `validate_checksum` public and have it **re-validate the checksum** and potentially turn the `Parsed<Ck>` into a `Parsed<OtherCk>`.
* We replace `parse_unvalidated` with special constructors for `Parsed<NoChecksum>` which skip checksum validation entirely, and make **these** public as well. (We'd add a unit test that if you used `new`/`new_with_witness_version` with `NoChecksum`, that you got the same result as these.)
* We clearly document that `new` and `new_with_witness_version` are the only constructors that norma users need to worry about.

Then in src/lib.rs, to deal with bech32 vs bech32m, we can take a `Parsed::new_with_witness_version_no_checksum(s)?` and then alternately call `revalidate_checksum::<Bech32>` or `revalidate_checksum::<Bech32m>` and save all the unnecessary iterations.

src/primitives/hrpstring.rs
```diff
@@ -0,0 +1,447 @@
...
+        self.data = &self.data[..data_len];
+
+        Ok(self)
+    }
+}
+
+/// A iterator over a parsed HRP string data as bytes.
+#[allow(clippy::type_complexity)]
+pub struct ByteIter<'s> {
+    // We need `Copied` because non of our adaptor iterators use referenced items.
```
apoelstra, line 160: In e2527dc19c3e8e9a4cd678a955fc5b161e175c87:

typo `non`. TBH you could probably delete the comment entirely, the purpose of `Copied` is pretty clear.

Also, I think `allow(clippy::type_complexity)` is probably the right approach here but FYI you can edit `type-complexity-threshold` in clippy.toml to just increase its threshold. The default is 250. In my experience 1000 shuts it up (and if it doesn't maybe you **should** break up the types :P).

src/primitives/hrpstring.rs
```diff
@@ -0,0 +1,447 @@
...
+    }
+}
+
+impl From<hrp::Error> for Error {
+    fn from(e: hrp::Error) -> Self { Error::InvalidHrp(e) }
+}
+
+impl From<Infallible> for Error {
+    fn from(v: Infallible) -> Self { match v {} }
+}
```
apoelstra, line 314: In e2527dc19c3e8e9a4cd678a955fc5b161e175c87:

I like this but I'm a bit curious where it comes from. We don't actually use `Infallible` anywhere. Is this a "best practice" thing or is it maybe a holdover from an earlier version of this project that actually did use it?

In the interest of API stability I'm tempted to drop these conversions and revisit them whenever `!` is stabilized.

src/primitives/iter.rs
```diff
@@ -0,0 +1,441 @@
...
+    /// Adapts the `Fe32` iterator to output bytes instead.
+    ///
+    /// If the total number of bits is not a multiple of 8, any trailing bits
+    /// are simply dropped.
+    fn fe32_to_byte_iter(mut self) -> Fe32ToByteIter<Self> {
+        Fe32ToByteIter { last_fe: self.next(), bit_offset: 0, iter: self }
+    }
+
+    /// Adapts the Fe32 iterator to append a checksum to the end of the data.
+    fn checksummed_iter<Ck: Checksum>(self) -> ChecksummedIter<Self, Ck> {
```
apoelstra, line 58: In e2527dc19c3e8e9a4cd678a955fc5b161e175c87:

These shouldn't have `iter` suffices if they're iterator adaptors.

apoelstra: I'd also make `fe32` and `byte` plural in both conversion adaptor names.

apoelstra: Or even do `fe32string_to_bytestring` and `bytestring_to_fe32string` to emphasize that it's not individual characters being converted.

src/primitives/iter.rs
```diff
@@ -0,0 +1,441 @@
...
+//! assert_eq!(hrpstring.to_uppercase(), "BC1QW508D6QEJXTDG4Y5R3ZARVARY0C5XW7KV8F3T4");
+//! ```
+
+use crate::primitives::checksum::{self, Checksum, PackedFe32};
+use crate::primitives::gf32::Fe32;
+use crate::primitives::hrp::{self, Hrp};
+
+/// Extension trait for byte iterators which provides an adaptor to GF32 elements.
+pub trait ByteIterExt: Sized + Iterator<Item = u8> {
+    /// Adapts the byte iterator to output GF32 field elements instead.
```
apoelstra, line 39: In e2527dc19c3e8e9a4cd678a955fc5b161e175c87:

Should document what happens if the total number of bits is not a multiple of 5. (0 bits are added.)

src/primitives/iter.rs
```diff
@@ -0,0 +1,441 @@
...
+}
+
+impl<I, Ck> ChecksummedIter<I, Ck>
+where
+    I: Iterator<Item = Fe32>,
+    Ck: Checksum,
+{
+    /// Creates a new checksummed iterator which adapts a data iterator of field elements by
+    /// appending the checksum.
+    pub fn new(data: I) -> ChecksummedIter<I, Ck> {
```
apoelstra, line 210: In e2527dc19c3e8e9a4cd678a955fc5b161e175c87:

Proposal:
* `ChecksummedIter::new` takes a HRP and a witness version.
* We add a separate `new_without_hrp` method if you're trying to use these adaptors with a non-hrpstring.
* If it has an HRP/witness byte, it should yield all the Fe32s that would be fed into the checksum engine
* Then a new `fn hrpstring` would adapt this to instead output characters (so, HRP chars instead of fe32s, etc)

apoelstra:
Review checkpoint e2527dc19c3e8e9a4cd678a955fc5b161e175c87

src/lib.rs
```diff
@@ -386,48 +161,30 @@ pub fn encode_to_fmt<T: AsRef<[u5]>>(
...
-                Err(e) => Ok(Err(e)),
-            }
-        }
+        Variant::Bech32 =>
+            for c in data
+                .as_ref()
+                .iter()
+                .copied()
+                .checksummed_iter::<Bech32>()
+                .hrpstring_iter(&hrp, None)
```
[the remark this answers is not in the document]
tcharding: Ah nice, I did this sort of thing in some test/fuzz code somewhere and its way nicer.

fuzz/fuzz_targets/encode_decode.rs
tcharding: Yes, you are correct. Used `expect` instead. Thanks

fuzz/fuzz_targets/encode_decode.rs
tcharding: We do that in `decode_rnd` although I'm not totally sure why we have both - do you see benefit in both of have I changed them so much that some previous benefit is gone? This test attempts to help the fuzzer find good addresses but with my new-found belief in the wonder magic of the fuzzer I'm not sure how useful that is?

src/lib.rs
tcharding: That's an optimization, right? I didn't put too much thought into how the crate API interacts with the new `primitives` code, except to ensure that we do not take any functionality away. I think we should tackle these questions when we re-write the whole API and/or add a segwit one.

src/primitives/hrpstring.rs
tcharding: Nice, used 1000 as suggested.

src/primitives/hrpstring.rs
```diff
@@ -0,0 +1,447 @@
...
+    /// The data payload is not a valid length.
+    InvalidDataLength,
+    /// The checksum does not match the rest of the data.
+    InvalidChecksum,
+    /// The checksum is not a valid length.
+    InvalidChecksumLength,
+    /// Some part of the string contains an invalid character.
+    InvalidChar(char),
+    /// The bit conversion failed due to a padding issue.
+    InvalidPadding,
```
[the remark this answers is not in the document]
tcharding: I wish there was a lint for unused error variants.

src/primitives/hrpstring.rs
tcharding: I can't remember why I added this, the compiler lead me to the original need for it and I copied the idea from some code Kix wrote somewhere IIRC. Its not used now so I've removed it.

src/primitives/iter.rs
tcharding: Cool, I elected for the following iterator types and methods that construct them
- `BytesToFes` and `bytes_to_fes`
- `FesToBytes` and `fes_to_bytes`
- `Checksummed` and `checksummed`
- `HrpstringChars` and `hrpstring_chars`

I feel like the last one could be better named, example usage:
```rust
    let iter = data.as_ref().iter().copied();
    match variant {
        Variant::Bech32 =>
            for c in iter.checksummed::<Bech32>().hrpstring_chars(&hrp, None) {
                fmt.write_char(c)?;
            },
        Variant::Bech32m =>
            for c in iter.checksummed::<Bech32m>().hrpstring_chars(&hrp, None) {
                fmt.write_char(c)?;
            },
    }
```

tcharding: And for more stimulation
```rust
let addr = witness_prog
    .iter()
    .copied()
    .bytes_to_fes()
    .checksummed::<Bech32>()
    .hrpstring_chars(&hrp, Some(witness_version))
    .collect::<String>()
```

src/primitives/iter.rs
tcharding: Yes, thanks. Done.

src/primitives/iter.rs
tcharding: Alright, this took a few iterations. Let me know what you think. I've kept at trying to enable various use cases while limiting the transitions from one adaptor to another to only valid transitions.

src/primitives/hrpstring.rs
tcharding: I did not get to processing this suggestion today - will come back to it.

tcharding:
Force push has all your suggestions except [0]. There was quite a bit of wild hacking today so I'd say we need to iterate again :)

[0] https://github.com/rust-bitcoin/rust-bech32/pull/113#discussion_r1245231234

fuzz/fuzz_targets/encode_decode.rs
apoelstra: Yes, there is value in both. One is checking that an arbitrary string can be decoded and re-encoded. The other is checking that an arbitarry hrpstring can be encoded and re-decoded.

src/primitives/iter.rs
apoelstra: All sounds good! I think `hrpstring_chars` is fine. `encoded` doesn't give any clue that the iterator adaptor is adding an HRP and a checksum.

apoelstra: Ah, I see, we've already called `hrp_checksummed` at this point. I think just `.chars()` is fine for this.

apoelstra: Actually I think we should invert this, and iterate over `char` by default, and require an adaptor to get the `fe`s out. We can call the latter `fes_for_checksum_engine`. Fine to have a mouthful of a name because it's not likely to be a commonly-used adaptor.

src/primitives/iter.rs
apoelstra: I like it! Had some comments but I stuck them on the other thread so I'll resolve this one.

src/primitives/hrpstring.rs
tcharding: I had a brief play with this. Using checksum generic on `Parsed` borks the current API in `libr.s` by making it impossible, for example, to return `Parsed` from `decode`. It also defeats the purpose of having the `Variant` type (which I do think will go away eventually but I don't think we should do that yet).

Can we leave this stuff until we start working out what the final crate API looks like? I"ll save a link to the stuff above so it doesn't get lost.

src/lib.rs
```diff
@@ -987,262 +303,101 @@ mod tests {
...
-
-        assert_eq!(encoded_str, written_str);
-    }
-
     #[test]
     #[cfg(feature = "alloc")]
     fn roundtrip_without_checksum() {
         let hrp = hrp("lnbc");
-        let data = "Hello World!".as_bytes().to_base32();
+        let data = "Hello World!".as_bytes().iter().copied().bytes_to_fes().collect::<Vec<u5>>();
```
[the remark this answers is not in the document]
tcharding: :) Used `bytes`, thanks.

src/primitives/hrpstring.rs
apoelstra: I'd really like to figure out how to use the API without doing twice as much work to parse segwit addresses as we used to.

apoelstra: Ok, we can punt on this until a later stage when we get rid of `Variant` in place of using the `Ck` parameter.

src/primitives/iter.rs
apoelstra: In 51781bc I'd still like to change this.

tcharding:
The `HrpstringChars::size_hint` function was incorrect. I noticed also that the `hrp` module has size_hint/len incorrect because of my previous lack of iterator knowledge, will fix as a separate PR.

apoelstra:
You can use `Option::zip` to make the `hrp_max`/`chk_max` stuff a little bit cleaner.

src/primitives/iter.rs
apoelstra: In d69ea01 I'd still like to change this.

src/lib.rs
tcharding: Oh by "re-write the whole API" I meant the public functions in `lib.rs` and if we add a segwit specific API (along with all the test vectors for bip 173/350), I have all this started in other branches from the last few months.

src/primitives/iter.rs
tcharding: hmm, interesting, just saw this after - will play with it now.

tcharding:
Another iteration on the API, this time I've added functions on the iterators that can be adapted instead of the extension trait. Also there are multiple transitions through all the iterators, see example in `iter` module.

tcharding:
And used `Option::zip` as suggested. Thanks, I've never used that before :)

src/primitives/iter.rs
```diff
@@ -0,0 +1,521 @@
...
+    /// if preset, and then appends a checksum.
+    fn hrp_checksummed<Ck: Checksum>(
+        self,
+        hrp: &Hrp,
+        witness_version: Option<Fe32>,
+    ) -> HrpChecksummed<Self, Ck> {
+        match witness_version {
+            Some(witness_version) =>
+                Checksummed::new(self).with_witness_version(witness_version).hrp_checksummed(hrp),
+            None => Checksummed::new(self).hrp_checksummed(hrp),
```
apoelstra, line 102: In caa79f50d9d74b99775dfcc14f66c4140e20397c:

Why call `Checksummed::new` here when the above function calls `self.checksummed`?

src/primitives/iter.rs
```diff
@@ -0,0 +1,521 @@
...
+    I: Iterator<Item = Fe32>,
+    Ck: Checksum,
+{
+    /// Adapts an checksummed iterator by inputting the HRP to the checksum algorithm.
+    ///
+    /// `iter` may or may not include the witness version.
+    pub fn new(hrp: &'hrp Hrp, mut iter: Checksummed<I, Ck>) -> HrpChecksummed<'hrp, I, Ck> {
+        iter.checksum_engine.input_hrp(hrp);
+        Self { hrp, iter }
+    }
```
apoelstra, line 353: In caa79f50d9d74b99775dfcc14f66c4140e20397c:

This `HrpChecksummed` doesn't actually output any of the HRP field elements. It needs to do that, otherwise the data you're outputting can't be validated or otherwise manipulated.

src/primitives/iter.rs
```diff
@@ -0,0 +1,521 @@
...
+    ) -> HrpChecksummed<Self, Ck> {
+        match witness_version {
+            Some(witness_version) =>
+                Checksummed::new(self).with_witness_version(witness_version).hrp_checksummed(hrp),
+            None => Checksummed::new(self).hrp_checksummed(hrp),
+        }
+    }
+
+    /// Adapts the Fe32 iterator to append a checksum to the end of the data.
+    fn checksummed<Ck: Checksum>(self) -> Checksummed<Self, Ck> { Checksummed::new(self) }
```
apoelstra, line 107: In caa79f50d9d74b99775dfcc14f66c4140e20397c:

It feels a little weird to have `hrp_checksummed` and also `checksummed`. It seems like maybe the HRP one should only prepend an HRP and not do anything else, so if you want both, you can do `.hrp(hrp, witness_version).checksummed()` which is the same number of characters but more flexible.

Though `.hrp` should probably be `.prepend_hrp`.

And I'd argue we should separate out the witness version and add a `.prepend_witness_version`.

src/primitives/iter.rs
tcharding: Good point, removed call to `new`.

src/primitives/iter.rs
tcharding: Cheers, I'll have another go.

src/primitives/iter.rs
tcharding: oh, that is different to what I had in mind. I'll have to think on that a bit.

tcharding:
To make review easier I opened #116 and then I attempted to extensively describe what this PR changes.

tcharding:
@clarkmoody this PR is pretty heavy and very invasive, if you have any concerns at all please voice them. Also if I can do anything else to make your review easier please say.

tcharding:
Holy molly, this is really hard to get a nice clean iterator API that conforms to the bips, is general purpose, but is sane. Converting this to draft while I keep playing with it.

tcharding:
@apoelstra I think I may not have it exactly clear what we are trying to achieve, I was unable yesterday to answer this question clearly to myself:

What is the aim of the iter API?

- to be fully flexible i.e, you can do what ever you want, transition from any iter to any other?
- to be concisely expressive for "typical" usecases?
- to be restrictive in which iterators can transition to which other iterators based on some set of usecases?

Sadly, the more I read bips 173/350 the more I am convinced it is not possible to write a correct, clean, general crate that can be used without reading those bips many times. I am still attempting however to do so and hopefully when we add a `segwit` module bitcoin users will get just such an API.

In #117 I've tried to make it easy and obvious how to use the API. Ergonomic for typical (from my limited understanding of typical) usecases but still fully general (I am having trouble working out all the general usecases).

tcharding:
#117 is definitely better than this one.

PR #114, opened 2023-06-29
Call through to inner size_hint

tcharding:
Currently our `size_hint` methods are calling `len` (from `ExactSizeIterator`) which is incorrect. We should in fact call `size_hint` on the inner iterator.

### DIFF
--- a/src/primitives/hrp.rs
+++ b/src/primitives/hrp.rs
@@ -207,7 +207,7 @@ pub struct ByteIter<'b> {
 impl<'b> Iterator for ByteIter<'b> {
     type Item = u8;
     fn next(&mut self) -> Option<u8> { self.iter.next().copied() }
-    fn size_hint(&self) -> (usize, Option<usize>) { (self.len(), Some(self.len())) }
+    fn size_hint(&self) -> (usize, Option<usize>) { self.iter.size_hint() }
 }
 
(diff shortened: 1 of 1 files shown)


apoelstra:
Was the original code actually wrong?

apoelstra:
The default implementation of `len` just passes through to `size_hint`. https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html

apoelstra:
Though those same docs also make the observation that when we have iterator adaptors that make things *longer*, we shouldn't be implementing `ExactSizeIterator` and we should be checking for overflow when computing our maximum size hint.

apoelstra (approved):
ACK 2a12d2e6a1a3f2c9c8ed74d1200f478f423c4c83

tcharding:
Sorry bro, I didn't respond before.

> Was the original code actually wrong?

I believe so because the default `len` calls `size_hint` so we introduced a potential cycle if someone removed the custom `len` implementation, which, now I read the docs again, seems like we should not have at all.

tcharding:
So I believe this PR is correct and I will do another PR to remove the `ExactSizeIterator` implementations.

clarkmoody (approved):
ACK 2a12d2e6a1a3f2c9c8ed74d1200f478f423c4c83

PR #115, opened 2023-07-08
Remove custom implementation of `ExactSizeIterator`

tcharding:
Draft because includes #114 (first 2 patches).

From the docs [0]

> The len method has a default implementation, so you usually shouldn’t
> implement it. However, you may be able to provide a more performant
> implementation than the default, so overriding it in this case makes
> sense.

Our current implementations do nothing but call through to the inner iterator's `len` method which is not a performance improvement so is contrary to the remarks above.

Remove the implementations of `ExactSizeIterator` using the default impl instead.

[0] https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html

### DIFF
--- a/src/primitives/hrp.rs
+++ b/src/primitives/hrp.rs
@@ -210,9 +210,7 @@ impl<'b> Iterator for ByteIter<'b> {
     fn size_hint(&self) -> (usize, Option<usize>) { self.iter.size_hint() }
 }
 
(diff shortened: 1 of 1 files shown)


apoelstra:
Calling into the inner iterator's `len` will be a performance improvement if the inner iterator has a more performant implementation than the default.

tcharding:
Oh touché.

PR #116, opened 2023-07-09
Do random improvements

tcharding:
Pull all the unrelated random improvements out of #113 so that we can review/merge them more easily. Nothing controversial here except perhaps commit: `a2209d5 Remove crate level examples`. Note also that commit: `1a746b7 Clean up comments in private error module` is new.

PR #117, opened 2023-07-12
Add support for encoding/decoding bech32 addresses using iterators

tcharding:
(Friendly note: clarkmoody unless you are just loving all this iterator stuff I suggest you don't bother reviewing this one (or #113) until apoelstra acks :) 

This is a slightly different approach to #113. Done on top of #116, so just the last patch. 

- Conversion to/from bytes/fes by way of the extension traits as in #113 (still in `iter.rs`)
- Checksum iter adapter as in #113 
- Decoding by way of three structs depending on usage.
- Encoding by way of a new struct `Encoder`, uses a builder-like pattern to construct the encoder and then one can call `chars()` to get a character iterator or `fes()` to get a field-element iterator (over all the fes that go into the checksum).  

This should mean:
- Cleaner API i.e., easier to use but also easier to not use incorrectly 
- Incorrect use of iterator adaptors should now be harder

IMNSHO the bech32 spec is convoluted as f**k, hiding all the details is not totally possible but this new API attempts to make it possible to correctly use the API without having to spend two weeks reading and re-reading bips 173 and 350.

### PR currently does not use the new API

For example usage see:
- the unit tests in `decode` and `encode`
- the module level rustdocs in `encode.rs` and `decode.rs`
- BIP-173 and BIP-350 test vectors in `tests/`

### DIFF
--- a/src/primitives/checksum.rs
+++ b/src/primitives/checksum.rs
@@ -97,12 +97,8 @@ impl<Ck: Checksum> Engine<Ck> {
 
     /// Feeds `hrp` into the checksum engine.
     pub fn input_hrp(&mut self, hrp: &Hrp) {
-        for b in hrp.lowercase_byte_iter() {
-            self.input_fe(Fe32(b >> 5));
-        }
-        self.input_fe(Fe32::Q);
-        for b in hrp.lowercase_byte_iter() {
-            self.input_fe(Fe32(b & 0x1f));
+        for fe in HrpFe32Iter::new(hrp) {
+            self.input_fe(fe)
         }
     }
 
@@ -200,3 +196,63 @@ macro_rules! impl_packed_fe32 {
 impl_packed_fe32!(u32);
 impl_packed_fe32!(u64);
 impl_packed_fe32!(u128);
+
+/// Iterator that yields the field elements that are input into a checksum algorithm for an [`Hrp`].
+pub struct HrpFe32Iter<'hrp> {
+    /// `None` once the hrp high fes have been yielded.
+    high_iter: Option<crate::hrp::LowercaseByteIter<'hrp>>,
+    /// `None` once the hrp low fes have been yielded.
+    low_iter: Option<crate::hrp::LowercaseByteIter<'hrp>>,
+}
+
+impl<'hrp> HrpFe32Iter<'hrp> {
+    /// Creates an iterator that yields the field elements of `hrp` as they are input into the
+    /// checksum algorithm.
+    pub fn new(hrp: &'hrp Hrp) -> Self {
+        let high_iter = hrp.lowercase_byte_iter();
+        let low_iter = hrp.lowercase_byte_iter();
+
+        Self { high_iter: Some(high_iter), low_iter: Some(low_iter) }
+    }
+}
+
+impl<'hrp> Iterator for HrpFe32Iter<'hrp> {
+    type Item = Fe32;
+    fn next(&mut self) -> Option<Fe32> {
+        if let Some(ref mut high_iter) = &mut self.high_iter {
+            match high_iter.next() {
+                Some(high) => return Some(Fe32(high >> 5)),
+                None => {
+                    self.high_iter = None;
+                    return Some(Fe32::Q);
+                }
+            }
+        }
+        if let Some(ref mut low_iter) = &mut self.low_iter {
+            match low_iter.next() {
+                Some(low) => return Some(Fe32(low & 0x1f)),
+                None => self.low_iter = None,
+            }
+        }
+        None
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let high = match &self.high_iter {
+            Some(high_iter) => {
+                let (min, max) = high_iter.size_hint();
+                (min + 1, max.map(|max| max + 1)) // +1 for the extra Q
+            }
+            None => (0, Some(0)),
+        };
+        let low = match &self.low_iter {
+            Some(low_iter) => low_iter.size_hint(),
+            None => (0, Some(0)),
+        };
+
+        let min = high.0 + 1 + low.0;
+        let max = high.1.zip(low.1).map(|(high, low)| high + 1 + low);
+
+        (min, max)
+    }
+}
--- a/src/primitives/encode.rs
+++ b/src/primitives/encode.rs
@@ -1,0 +1,332 @@
+// SPDX-License-Identifier: MIT
+
+//! Bech32 address encoding.
+//!
+//! This module provides types and iterators that can be used to encode data as a bech32 address in
+//! a variety of ways without any allocations, generating, verifying, and appending checksums,
+//! prepending HRP strings etc.
+//!
+//! In general, directly using these adaptors is not very ergonomic, and users are recommended to
+//! instead use the higher-level functions at the root of this crate.
+//!
+//! # Examples
+//!
+//! ```
+//! use bech32::{Bech32, ByteIterExt, Fe32IterExt, Fe32, Hrp};
+//!
+//! let witness_prog = [
+//!     0x75, 0x1e, 0x76, 0xe8, 0x19, 0x91, 0x96, 0xd4,
+//!     0x54, 0x94, 0x1c, 0x45, 0xd1, 0xb3, 0xa3, 0x23,
+//!     0xf1, 0x43, 0x3b, 0xd6,
+//! ];
+//!
+//! // Get a stream of characters representing the bech32 encoded
+//! // address using "bc" for the human-readable part.
+//! let hrp = Hrp::parse("bc").expect("bc is valid hrp string");
+//! let chars = witness_prog
+//!     .iter()
+//!     .copied()
+//!     .bytes_to_fes()
+//!     .with_checksum::<Bech32>(&hrp)
+//!     .with_witness_version(Fe32::Q) // Optionally add witness version.
+//!     .chars();
+//!
+//! #[cfg(feature = "alloc")]
+//! {
+//!     let addr = chars.collect::<String>();
+//!     assert_eq!(addr.to_uppercase(), "BC1QW508D6QEJXTDG4Y5R3ZARVARY0C5XW7KV8F3T4");
+//! }
+//! ```
+
+use core::iter::Iterator;
+use core::marker::PhantomData;
+
+use crate::primitives::checksum::HrpFe32Iter;
+use crate::primitives::hrp::{self, Hrp};
+use crate::primitives::iter::Checksummed;
+use crate::{Checksum, Fe32};
+
+/// The `Encoder` builds iterators that can be used to encode field elements into a bech32 address.
+///
+/// Construct the encoder by calling [`Fe32IterExt::with_checksum`] on an iterator of field
+/// elements, optionally prefix the data with a witness version, and then get the encoding as either
+/// a stream of characters ([`Encoder::chars`]) or a stream of field elements ([`Encoder::fes`]).
+///
+/// # Examples
+///
+/// ```
+/// use bech32::{Bech32, ByteIterExt, Fe32IterExt, Hrp};
+///
+/// let data = [0x75, 0x1e, 0x76, 0xe8, 0x19, 0x91, 0x96, 0xd4];
+///
+/// let hrp = Hrp::parse("abc").expect("bc is valid hrp string");
+/// let chars = data
+///     .iter()
+///     .copied()
+///     .bytes_to_fes()
+///     .with_checksum::<Bech32>(&hrp)
+///     .chars();
+/// ```
+/// [`Fe32IterExt::with_checksum`]: crate::Fe32IterExt::with_checksum
+#[derive(Clone, PartialEq, Eq)]
+pub struct Encoder<'hrp, I, Ck>
+where
+    I: Iterator<Item = Fe32>,
+    Ck: Checksum,
+{
+    /// The field elements to encode.
+    data: I,
+    /// The human-readable part used at the front of the address encoding.
+    hrp: &'hrp Hrp,
+    /// The witness version, if present.
+    witness_version: Option<Fe32>,
+    /// Checksum marker.
+    marker: PhantomData<Ck>,
+}
+
+impl<'hrp, I, Ck> Encoder<'hrp, I, Ck>
+where
+    I: Iterator<Item = Fe32>,
+    Ck: Checksum,
+{
+    /// Constructs a new bech32 encoder.
+    pub fn new(data: I, hrp: &'hrp Hrp) -> Self {
+        Self { data, hrp, witness_version: None, marker: PhantomData::<Ck> }
+    }
+
+    /// Adds `witness_version` to the encoder (as first byte of encoded data).
+    ///
+    /// Note, caller to guarantee that witness version is within valid range (0-16).
+    pub fn with_witness_version(mut self, witness_version: Fe32) -> Self {
+        self.witness_version = Some(witness_version);
+        self
+    }
+
+    /// Returns an iterator that yields the bech32 encoded address as field ASCII characters.
+    pub fn chars(self) -> CharIter<'hrp, I, Ck> {
+        let witver_iter = WitnessVersionIter::new(self.witness_version, self.data);
+        CharIter::new(self.hrp, witver_iter)
+    }
+
+    /// Returns an iterator that yields the field elements that go into the checksum, as well as the checksum at the end.
+    ///
+    /// Each field element yielded has been input into the checksum algorithm (including the HRP as it is fed into the algorithm).
+    pub fn fes(self) -> Fe32Iter<'hrp, I, Ck> {
+        let witver_iter = WitnessVersionIter::new(self.witness_version, self.data);
+        Fe32Iter::new(self.hrp, witver_iter)
+    }
+}
+
+/// Iterator adaptor that just prepends a single character to a field element stream.
+///
+/// More ergonomic to use than `std::iter::once(fe).chain(iter)`.
+pub struct WitnessVersionIter<I>
+where
+    I: Iterator<Item = Fe32>,
+{
+    witness_version: Option<Fe32>,
+    iter: I,
+}
+
+impl<I> WitnessVersionIter<I>
+where
+    I: Iterator<Item = Fe32>,
+{
+    /// Creates a [`WitnessVersionIter`].
+    pub fn new(witness_version: Option<Fe32>, iter: I) -> Self { Self { witness_version, iter } }
+}
+
+impl<I> Iterator for WitnessVersionIter<I>
+where
+    I: Iterator<Item = Fe32>,
+{
+    type Item = Fe32;
+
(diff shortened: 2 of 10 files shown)


src/primitives/checksum.rs
```diff
@@ -200,3 +196,63 @@ macro_rules! impl_packed_fe32 {
 impl_packed_fe32!(u32);
 impl_packed_fe32!(u64);
 impl_packed_fe32!(u128);
+
+/// Iterator that yields the field elements that are input into a checksum algorithm for an [`Hrp`].
+pub struct HrpChecksumFe32Iter<'hrp> {
```
apoelstra, line 201: In 5c9b549a3738cec82f633c14e274854b82edce45:

I'd drop `Checksum` from this name. I don't think it really clarifies anything.

src/primitives/decode/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+///
+/// Pre-parsing an HRP string only checks validity of the characters, it does not validate the
+/// checksum in any way - to validate convert to [`Parsed`].
+#[derive(Debug)]
+pub struct Parsed<'s> {
+    /// The human-readable part, guaranteed to be lowercase ASCII characters.
+    pub(crate) hrp: Hrp,
+    /// This is ASCII byte values of the parsed string, guaranteed to be valid bech32 characters,
+    /// with the checksum removed.
+    pub(crate) data: &'s [u8],
```
apoelstra, line 47: In 5c9b549a3738cec82f633c14e274854b82edce45:

I'd drop `pub(crate)` from both of these, and add a public accessor to `data`.

src/primitives/decode/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+        Fe32Iter { iter: AsciiToFe32Iter { iter: self.data.iter().copied() } }
+    }
+
+    /// Parses an bech32 encode string and constructs a [`Parsed`] object.
+    ///
+    /// Checks for valid ASCII values, does not validate the checksum.
+    ///
+    /// The object returned by this function must must have `validate_checksum` called on it to
+    /// validate and checksum and maintain data invariant (remove checksum from parsed data).
+    fn parse_unvalidated(s: &'s str) -> Result<Self, Error> {
```
apoelstra, line 79: In 5c9b549a3738cec82f633c14e274854b82edce45:

I'd call this `new_unvalidated` to match `new`.

apoelstra: Oh, I see, it's not `pub`. No need to make change the doccomment then. Or the name, arguably.

src/primitives/decode/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+    type Item = Fe32;
+    fn next(&mut self) -> Option<Fe32> { self.iter.next() }
+    fn size_hint(&self) -> (usize, Option<usize>) { self.iter.size_hint() }
+}
+
+/// Helper iterator adaptor that maps an iterator of valid bech32 character ASCII bytes to an
+/// iterator of field elements.
+///
+/// This iterator is a performance optimization. Equivalent, but significantly faster than, using
+/// `hrp_string.data_chk.iter().copied().map(|c| Fe32::from_char_unchecked(c).to_u8())`.
```
apoelstra, line 150: In 5c9b549a3738cec82f633c14e274854b82edce45:

This comment is out of date and can probably just be dropped. There is no `data_chk` field on `Parsed`.

src/primitives/decode/segwit.rs
```diff
@@ -0,0 +1,112 @@
...
+    /// Returns the human-readable part.
+    pub fn hrp(&self) -> Hrp { self.inner.hrp() }
+
+    /// Returns an iterator over the byte data encoded by the HRP string (excluding the HRP, witness
+    /// version byte if any, and checksum).
+    pub fn byte_iter(&self) -> ByteIter { self.inner.byte_iter() }
+
+    /// Returns an iterate over field elements of the data encoded by the HRP string (excluding the
+    /// HRP, witness version byte if any, and checksum).
+    pub fn fe32_iter(&self) -> Fe32Iter { self.inner.fe32_iter() }
```
apoelstra, line 60: In 5c9b549a3738cec82f633c14e274854b82edce45:

I think this should be dropped. It's a little confusing to have an iterator over field elements that doesn't include the HRP data or checksum, and it's not clear to me what the value is. Note that this is not used anywhere.

src/primitives/decode/mod.rs
```diff
@@ -0,0 +1,384 @@
...
+
+/// Separator between the hrp and payload (as defined by BIP-173).
+const SEP: char = '1';
+
+/// An HRP string that has been parsed and had the checksum validated.
+///
+/// Pre-parsing an HRP string only checks validity of the characters, it does not validate the
+/// checksum in any way - to validate convert to [`Parsed`].
+#[derive(Debug)]
+pub struct Parsed<'s> {
```
apoelstra, line 42: In 5c9b549a3738cec82f633c14e274854b82edce45:

This (and the other struct) should be renamed. Now that they are no longer in a module called `hrpstring` the name `Parsed` does not make sense.

src/primitives/decode/segwit.rs
apoelstra: Same with the other one, on the othe `Parsed`.

src/primitives/decode/segwit.rs
```diff
@@ -0,0 +1,112 @@
...
+
+    /// Returns the witness version if present.
+    pub fn witness_version(&self) -> u8 { self.witness_version }
+
+    /// Returns the human-readable part.
+    pub fn hrp(&self) -> Hrp { self.inner.hrp() }
+
+    /// Returns an iterator over the byte data encoded by the HRP string (excluding the HRP, witness
+    /// version byte if any, and checksum).
+    pub fn byte_iter(&self) -> ByteIter { self.inner.byte_iter() }
```
apoelstra, line 56: In 5c9b549a3738cec82f633c14e274854b82edce45:

This says that it excludes the witness version byte, but I don't see where in the code this is actually excluded. Also none  of the witness byte comments here should say "if any" since there is always a witness version byte.

src/primitives/decode/segwit.rs
```diff
@@ -0,0 +1,112 @@
...
+    witness_version: u8,
+}
+
+impl<'s> Parsed<'s> {
+    /// Parses an HRP string, treating the first data character as a witness version.
+    ///
+    /// This version byte does not appear in the extracted binary data, but is covered
+    /// by the checksum. It can be accessed with [`Self::witness_version`] and is also
+    /// returned from this constructor as a convenience.
+    pub fn new<Ck: Checksum>(s: &'s str) -> Result<Self, Error> {
```
apoelstra, line 31: In 5c9b549a3738cec82f633c14e274854b82edce45:

We might as well do the bech32/bech32m decision here rather than making the user choose a checksum (which they cannot do until they've learned the witness version).

src/primitives/encode.rs
```diff
@@ -0,0 +1,311 @@
...
+{
+    /// Constructs a new bech32 encoder.
+    pub fn new(data: I, hrp: &'hrp Hrp) -> Self {
+        Self { data, hrp, witness_version: None, marker: PhantomData::<Ck> }
+    }
+
+    /// Add `witness_version` to the encoder (as first byte of encoded data).
+    ///
+    /// Note, caller to guarantee that witness version is within valid range (0-16).
+    pub fn prepend_witness_version(mut self, witness_version: Fe32) -> Self {
```
apoelstra, line 79: In 5c9b549a3738cec82f633c14e274854b82edce45:

I'd panic here if `self.witness_version` is already set. Or just rename this method to `with_witness_version` or something.

src/primitives/iter.rs
```diff
@@ -0,0 +1,473 @@
...
+    /// Adapts the `Fe32` iterator to output bytes instead.
+    ///
+    /// If the total number of bits is not a multiple of 8, any trailing bits
+    /// are simply dropped.
+    fn fes_to_bytes(mut self) -> FesToBytes<Self> {
+        FesToBytes { last_fe: self.next(), bit_offset: 0, iter: self }
+    }
+
+    /// Adapts the Fe32 iterator to encode the field elements into a bech32 address.
+    fn encoder<Ck: Checksum>(self, hrp: &Hrp) -> Encoder<Self, Ck> { Encoder::new(self, hrp) }
```
apoelstra, line 69: In 5c9b549a3738cec82f633c14e274854b82edce45:

I'd rename `encoder` to `with_checksum`

apoelstra: I guess `with_checksum_hrp`. And we could add a `with_checksum` method which just creates a `Checksummed`, no HRP.

apoelstra:
Done reviewing 5c9b549a3738cec82f633c14e274854b82edce45. This is definitely shaping up.

src/primitives/iter.rs
tcharding: I was trying to show from the name that its not an iterator adapter but a builder.

tcharding:
Thanks for the review @apoelstra, I read your comments but didn't get to work on this today. I'm good to implement all suggestions except unsure about the thing I commented on.

src/primitives/iter.rs
apoelstra: I'd argue that it's still an iterator adaptor, because it's a method on an arbitrary iterator. I get your point that it's not an iterator by itself, but I think it's more confusing/less discoverable to give it a name like `encoder`, which could mean pretty-much anything.

tcharding: Fair points, I'll have another play with it.

src/primitives/checksum.rs
tcharding: Nice, done.

src/primitives/decode/mod.rs
tcharding: Woops, we don't even access the data anymore. I left them private for now.

src/primitives/decode/mod.rs
tcharding: Removed.

src/primitives/decode/segwit.rs
tcharding: Awesome, this prompted me to add all the test vector integration tests I wrote ages ago to the PR!

src/primitives/decode/segwit.rs
tcharding: My bad, thanks.

src/primitives/decode/mod.rs
tcharding: Renamed them to `Hrpstring`.

src/primitives/decode/mod.rs
tcharding: I've re-worked the constructors to get the test vectors all running - includes re-naming and making functions public.

src/primitives/decode/segwit.rs
tcharding: Dropped!

src/primitives/iter.rs
tcharding: Changed the `encoder` function to be called `with_checksum` - FTR I don't really like it but I cannot come up with anything better or reasonable argument against your points :)

tcharding:
I pushed up what I've done. Currently there is no invariant on the checksum being in `data` (`decode::Hrpstring`).  Perhaps we should pull out a `Parsed` type again from `decode::Hrpstring` that is the pre-parsed concept we had before. I had a play but didn't get anything I really liked.

tcharding:
ooo, I've come up with something mad. Converting to draft while I polish it, will push shortly.

tcharding:
Hey @apoelstra check this re-write of the `decode` module out. I'm not happy with the three type names but I'm super happy with the abstractions - let me know what you think. (Its in a separate patch, will squash once you've had a look.)

apoelstra:
@tcharding looks great! I'm unsure whether `KnownHrp` is worth the reduction in flexibility, since we don't really use it for anything.

Agreed that the names are no good. I suggest
* `Parsed` becomes `UncheckedHrpString` or maybe `Unchecked`
* `Base32` becomes `HrpString`
* `Segwit` can stay as `Segwit`, or maybe it should be `SegwitHrpString`.

tcharding:
I dug some more into the whole hrp <-> network thing and came up with something different, added a module called `network_kind` and functions `is_valid_on_mainnet` (on `Segwit` and `Hrp`). I haven't tried using the new bech32 in bitcoin for a while, perhaps I'll try and see if this new stuff adds any vaule.

tcharding:
Used `UnvalidatedHrpString` instead so as not to overload "check" and `SegwitHrpString`, I think we should err on the side of explicitness with this module.

tcharding:
Removed `KnownHrp` and inlined the check.

tcharding:
Just the padding thing left to do.

tcharding:
Ok, I changed two of the validate functions that convert from one type to another `HrpString` to `SegwitHrpString` to consume self, can you think of any reason why this might be too restrictive?

I found some bugs in the `len` methods I wrote, now fixed.

tcharding:
Ok, i'll stop polishing it now :)

apoelstra:
`UnvalidateHrpString` sounds wrong because we do validate that it's an HRP string: it has an HRP, the character set is correct, etc. We just don't check any checksums. What word would you use rather than "check" for a checksum?

apoelstra (approved):
ACK 3a1eca84b8abca734745e96f792898e68f3e8cfb

tcharding:
Yeah I agree with everything you said, its really hard to write this so that it can be used as the bip says and be actually useful in real life and easy to use. The bips are really hard to write a clean API against. I'll have another go today.

tcharding:
Ill try and get it to a state where its only naming that is borked, then perhaps we can move on to writing the higher level api and see if bettter name fall out. Also I think that basically no-one should ever use this decode API directly, perhaps I'll add a big comment saying that if one is planning on using this module they should expect to spend a bunch of time reading the bips, if they do not want to do that then do not use the module, its too convoluted.

apoelstra:
Lol, well, if you avoid `SegwitHrpString` then I think it's pretty sensible.

tcharding:
On the drive home from school drop off just now I was thinking about this, I'm not keen on the `HrpString` bit because all three are hrpstrings (as you say, guaranteed to have valid bech32 characters etc.). When writing docs, and still in there, I kept using the word "parsed" for the initial stage of parsing the string and checking the characters - hence `Parsed`. The other two are exactly designed to mirror the bip statement 

> We first describe the general checksummed base32 format called Bech32 and then define
> Segregated Witness addresses using it.

What I tried to model before using the modules and I tried to use names that were terse but matched that sentence - hence `Base32` and `Segwit`. But again we are back to a place where one has to know the bips inside out to use the API.

apoelstra:
> * `UncheckedHrpString`, `CheckedHrpString`, and `SegwitHrpString`

I like this one. As you say, in this module it's better to be explicit, since we don't expect it to be directly used much. So it's okay to have the verbose/repetitive suffixes.

Another thing we should bikeshed is whether `Hrpstring` is one word or not -- if so, the `S` should be lowercase. If not, we should stick a `_` into all the snake case versions of it. My vote is for `Hrpstring`.

tcharding:
Oh woops, yes I had settled on `hrpstring` in local variables but, well observed, I messed up using `HrpString` - will change to `Hrpstring`.

tcharding:
> I don't really like these "valid HRP" checks, but I guess they're necessary to pass the test vectors in the BIP, so whatever. But it's an extra error variant and codepath to manage that provides zero value because nobody ever wants a "HRP string where the HRP is bc or tb, but I don't care which".

Yeah I toyed briefly with the idea of adding separate functions that include this hrp stuff and then not bothering with it for the "normal" code path.

- `CheckedHrpstring::validate_segwit_strict`
- `SegwitHrpstring::new_strict`
- `SegwitHrpstring::new_bech32_strict`

I had a quick go at implementing this, its ugly.
-

tcharding:
Changes in force push:
- Added a bunch more rustdocs incl. examples
- Removed the `with_checksum` function from `BytesIterExt` (I added it recently and it is no good).

apoelstra (approved):
ACK 3ca7a190f0cd1acd287814b3e2a838e2f0fc9923

tcharding:
Changes in force push (all to `decode`:
- Fixed stale local variable names (after all the type naming churn)
- Fixed an error variant (also stale)
- Fixed incorrect docs on error types (cut'n'pasta fails)

tcharding:
@clarkmoody this one is ready for review if you have the cycles to spare please. Take your time.

tcharding:
While depending on this branch in `rust-bitcoin` I noticed one error variant was misnamed, renamed `SegwitError` -> `SegwitHrpstringError`

tcharding:
Ah sorry @clarkmoody, might want to wait some more ...

@apoelstra this HRP ("bc", "tb", "bcrt") thing is pernicious. I attempted to use this branch today in `rust-bitcoin` and its basically unusable because of the HRP checks, so one cannot parse a get any segwit checks done. Push a quick Friday afternoon fix to get the discussion started.

apoelstra:
ACK e3d62d4bd11eba39d801e40ac4546bf63e74d9d1 except that the last commit should be cleaned up (at least, fix the CI message and remove 'WIP' from the commit message).

tcharding:
Cleaned up and squashed! Note also the additional functions on `Hrp` commented on above.

apoelstra (approved):
ACK de7da50ad3c57dd39f6753fe1f708e8b80f18be3

tcharding:
Ok @clarkmoody, now its ready for you please if you have time.

tcharding:
If/when this merges should we consider releasing a new version at this stage? The benefit is that the `primitives` module is in place so if we release we can use it in `rust-bitcoin` and the play around with it and improve it as needed. Also I do not think that we can get much more done in time for the next release of `rust-bitcoin` and it would be super nice to be using this stuff already so we can start work on other address improvements in `rust-bitcoin`. Is there any downsides of doing so that I'm overlooking?

apoelstra:
@tcharding we should look at the diff and see how many APIs we've deleted already, and consider trying to make best-effort replacements of those as a stop-gap.

apoelstra:
Actually, I think we should release this with a `-alpha0` suffix or something rather than trying to emulate the old API. Then people can use it but nobody will feel forced into it.

tcharding:
> I think we should release this with a -alpha0 suffix

Good idea.

tcharding:
Pushed code comment change to bip-173 test to mirror the bip-350 one.

apoelstra (approved):
ACK 5478e2bd307991557e006e294f528c94b52fd6ff

tcharding:
If you are interested @clarkmoody there is a PR in `rust-bitcoin` showcasing usage of this branch.

https://github.com/rust-bitcoin/rust-bitcoin/pull/1951

clarkmoody (approved):
Massive changes! Thanks for the hard work. Agreed with the comment in `rust-bitcoin` about a shortcut method to collapse that large iterator chain.

tcharding:
Thanks man, we worked hard on this one.

PR #121, opened 2023-08-10
primitives: Inline public functions

tcharding:
Audit the `primitives` module and all submodules; add inline to all public methods that are small enough (excl. error impls).

### DIFF
--- a/src/primitives/checksum.rs
+++ b/src/primitives/checksum.rs
@@ -93,9 +93,11 @@ impl<Ck: Checksum> Default for Engine<Ck> {
 
 impl<Ck: Checksum> Engine<Ck> {
     /// Constructs a new checksum engine with no data input.
+    #[inline]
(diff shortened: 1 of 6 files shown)


apoelstra:
concept ACK. I'm not sure what your exact criteria is and I'm not sure how to vet your choices, but nothing here seems particularly egregious.

If we wanted a specific policy I think "one-liners and functions we've actually benchmarked to be faster with inlining" would be a good one.

apoelstra (approved):
ACK 847e1c97ae6745fd38d21f049d602a15dd030d7e

tcharding:
> I'm not sure what your exact criteria is 

My critera was, after re-reading #1595 discussion, "kind of guess at what the assembler would be, given I know very little about compiler optimisations, and then aggressively inline".

clarkmoody (approved):
ACK 847e1c97ae6745fd38d21f049d602a15dd030d7e

Issue #19, opened 2018-04-15
Implement CheckBase32 more efficiently for slices


sgeisler:
#17 introduces the `CheckBase32` trait which allows to check if all values of a `[u8]` are in the range `0..32` and returns the values of the slice as `T: AsRef<[u5]>`. Currently this can only be implemented for `T=Vec<u5>` because of uncertainties of the memory layout of `u5` ([see discussion](https://www.reddit.com/r/rust/comments/88xvgg/hey_rustaceans_got_an_easy_question_ask_here/dwvk0w9/)). Hopefully rust-lang/rust#43036 will resolve this uncertainty.

If `[u8]` and `[u5]` would be guaranteed to have the same memory layout we could do a simple cast from `&'f [u8]` to `&'f [u5]` after checking that all values are in range and save an allocation.

clarkmoody:
@sgeisler Is this issue still relevant. The code has changed a bit since April of '18 😄

sgeisler:
I guess not. The only place where the trait is used in the library itself is in tests and the only thing it would be useful for otherwise is communicating with bech32 but not u5 aware libraries (which don't exist afaik since it's so easy to just get the u5 type from this lib). Although the language feature I referenced is now stable it would probably take an eternity until our oldest supported rust version can use it. So I guess we can close it, if anyone feels a need for it in the future it can be reopened.

PR #122, opened 2023-08-11
Add transparent representation for u5

citizen-stig:
As mentioned in https://github.com/rust-bitcoin/rust-bech32/issues/19 having transparent representation for u5 opens many optimization.

According to change log in Rust repo, this feature was [stabilized in 1.28](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1280-2018-08-02)


apoelstra:
LGTM. I don't see your citation for this opening up optimizations, but it seems like an obvious thing to do (and we'll never change this type from being a u8 internally).

apoelstra (approved):
ACK 02a5686db576361c8a21ebe8d4bc054ca5b093ac

PR #125, opened 2023-08-21
Remove "the rust bitcoin developers" from manifest

tcharding:
We don't use this phrase much anymore, remove it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "bech32"
 version = "0.10.0-alpha"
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK df67db91b5590095b3580baa85ae98457dad5a99

apoelstra:
Fine by me. Will let @clarkmoody review as well. `git log | grep 'Author:' | sort | uniq -c` shows that there are not many other authors, and of the ones there are, most are not even rust-bitcoin regulars.

The only other person with more than one PR is @sgeisler who is no longer active.

clarkmoody (approved):
I'm fine either way

Issue #126, opened 2023-08-25
0.10-alpha API changes


apoelstra:
You can see the documentation for 0.10 at docs.rs: https://docs.rs/bech32/0.10.0-alpha/bech32/index.html

A few comments:

* We have `FromBase32`, `ToBase32`, `WriteBase32`, `WriteBase256`, `CheckBase32` traits. None of these are used anywhere. I think every one of these traits should be dropped.
* If you have an allocator, the `encode` and `decode` functions (and their variants) are available. We should draw attention to these in the main docs. If you want segwit, we should point the user to the `segwit` module.
* If you do *not* have an allocator, we currently have the messy arrayvec-based API with functions like `decode_lowercase` which really aren't useful in the absense of arrayvec.

I propose:

* We drop all the existing traits.
* We expose `primitives::decode::CheckedHrpString` at the top level for non-alloc decoding
* We expose a wrapper for the `data.iter().copied().bytes_to_fes().with_checksum::<Bech32>(&hrp).chars()` iterator chain, and maybe another for the `with_witness_version` variant, at the top level for non-alloc encoding. These wrappers should just take a byte slice. We can document that if you want to use an arbitrary iterator, the `primitives::iter` module will give you the pieces to do so.
* We rewrite the top-level docs to give examples of various things:
   * A normal Bitcoin user probably wants to use the `segwit` module; see examples there.
   * A non-Bitcoin user probably wants to use `encode`/`decode` functions (add example).
   * A non-Bitcoin no-alloc user probably wants to use the `CheckedHrpString` and non-alloc decoding method.
   * If you want to define your own checksum you should use the `Checksum` trait. We should provide an example, maybe of the BIP-93 checksum.
* We should also change all the re-exports to use `#[doc(inline)]` so that they appear as top-level structures rather than re-exports.

tcharding:
Perhaps we can come up with a list of the users we are trying to support because I don't currently know this? The reason I ask is that removing all the `lib.rs` traits basically makes the current top level API useless because it's tightly coupled eg, the `encode` function takes `T: AsRef[u5]` because it was expected that users would use the `ToBase32` trait to convert their `Vec<u8>` to a `&[u5]`.

tcharding:
[...]

PR #128, opened 2023-09-04
Re-write crate level API

tcharding:
Now includes updates to fuzzing and is on top of #130 (bug fix).

Now we have all the new `primitives` pieces in place we can re-write the public API to take advantage of them - WIN!

Close: #126

### DIFF
--- a/embedded/no-allocator/src/main.rs
+++ b/embedded/no-allocator/src/main.rs
@@ -7,9 +7,9 @@
 #![no_main]
 #![no_std]
 
-use arrayvec::{ArrayString, ArrayVec};
-use bech32::{self, u5, Hrp, Variant, ByteIterExt, Bech32};
+use arrayvec::ArrayString;
 use bech32::primitives::decode::CheckedHrpstring;
+use bech32::{Bech32, Hrp};
 use cortex_m_rt::entry;
 use cortex_m_semihosting::{debug, hprintln};
 use panic_halt as _;
@@ -20,20 +20,21 @@ use panic_halt as _;
 fn main() -> ! {
     let mut encoded = ArrayString::<30>::new();
 
-    let base32 = [0x00u8, 0x01, 0x02].iter().copied().bytes_to_fes().collect::<ArrayVec<u5, 30>>();
+    let data = [0x00u8, 0x01, 0x02];
+    let hrp = Hrp::parse("bech32").expect("failed to parse hrp");
 
-    let hrp = Hrp::parse("bech32").unwrap();
-
-    bech32::encode_to_fmt_anycase(&mut encoded, hrp, &base32, Variant::Bech32).unwrap().unwrap();
+    bech32::encode_to_fmt::<Bech32, _>(&mut encoded, hrp, &data)
+        .expect("failed to encode");
     test(&*encoded == "bech321qqqsyrhqy2a");
 
     hprintln!("{}", encoded).unwrap();
 
-    let unchecked = CheckedHrpstring::new::<Bech32>(&encoded).unwrap();
+    let unchecked =
+        CheckedHrpstring::new::<Bech32>(&encoded).expect("failed to construct CheckedHrpstring");
+    let iter = unchecked.byte_iter();
 
     test(unchecked.hrp() == hrp);
-    let res = unchecked.byte_iter().collect::<ArrayVec<u8, 30>>();
-    test(&res == [0x00, 0x01, 0x02].as_ref());
+    test(iter.eq(data.iter().map(|&b| b)));
 
     debug::exit(debug::EXIT_SUCCESS);
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9,8 +9,110 @@
 //! a data part. A checksum at the end of the string provides error detection to prevent mistakes
 //! when the string is written off or read out loud.
 //!
+//! # Usage
+//!
+//! - If you are doing segwit stuff you likely want to use the [`segwit`] API.
+//! - Non-segwit stuff and you have an allocator, use the top level API. For normal usage the
+//!   [`encode`] and [`decode`] functions should suffice. There are also various other functions for
+//!   explicit control of the checksum algorithm and the case used when encoding.
+//! - Non-segwit stuff and you do *not* have an allocator, use the
+//!   [`primitives::decode::CheckedHrpstring`] type for decoding. For encoding we provide various
+//!   top level functions of the form `encode*_to_fmt`.
+//! - To define your own checksum algorithm implement [`crate::Checksum`] (see example below).
+//!
 //! The original description in [BIP-0173](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki)
 //! has more details. See also [BIP-0350](https://github.com/bitcoin/bips/blob/master/bip-0350.mediawiki).
+//!
+//! # Examples
+//!
+//! ## Encoding
+//!
+//! ```
+//! # #[cfg(feature = "alloc")] {
+//! use bech32::{hrp, segwit, Hrp, Bech32m};
+//!
+//! const DATA: [u8; 20] = [0xab; 20]; // Arbitrary data to be encoded.
+//! const ADDR: &str = "abc14w46h2at4w46h2at4w46h2at4w46h2at958ngu";
+//! const TAP_ADDR: &str = "bc1p4w46h2at4w46h2at4w46h2at4w46h2at5kreae";
+//!
+//! // Encode arbitrary data using "abc" as the human-readable part and append a bech32m checksum.
+//! let hrp = Hrp::parse("abc").expect("valid hrp");
+//! let address = bech32::encode::<Bech32m>(hrp, &DATA).expect("failed to encode address");
+//! assert_eq!(address, ADDR);
+//!
+//! // Encode arbitrary data as a Bitcoin taproot address.
+//! let taproot_address = segwit::encode(&hrp::BC, segwit::VERSION_1, &DATA).expect("valid witness version and program");
+//! assert_eq!(taproot_address, TAP_ADDR);
+//!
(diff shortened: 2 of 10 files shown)


tcharding:
This PR more or less follows the proposal in #126 but I'm not auto-closing it because it does not include the suggested checksum example.

> If you want to define your own checksum you should use the Checksum trait. We should provide an example, maybe of the BIP-93 checksum.

tcharding:
This might need splitting up into separate PRs now because I've made some changes to the HRP constants to make them more readable.

I've audited the two top level APIs (lib.rs and segwit.rs) and I'm happy they are uniform. Open to naming improvements still of course.

I've written docs and examples, still TODO is the custom checksum doc.

Note please the new `crate::hrp` module that is soley to improve import ergonomics because I noticed we use `hrp::BC` any time we use the segwit API but one used to have to reach into `primitives` to get it.

tcharding:
Ok, if CI goes green this is ready for review please @apoelstra, I edited the "what I did" post above, so in case you read a stale version please not this now adds `Codex32`. If you are more-or-less happy with the current state I can spilt the PR up. I'll ping you @clarkmoody when we are at a "final" re-viewable state so you don't have to follow along if you don't want to.

tcharding:
FYI to check the embedded tests locally one must pin with `cargo update -p proc-macro2 --precise 1.0.60`. I can't remember how to run the crate, using `run` causes a core dump.

tcharding:
In case its useful to you @clarkmoody I'm building the docs locally to review them using

`RUSTDOCFLAGS="--cfg docsrs" cargo +nightly rustdoc --all-features -- -D rustdoc::broken-intra-doc-links`

embedded/no-allocator/src/main.rs
```diff
@@ -20,20 +20,21 @@ use panic_halt as _;
...
     let mut encoded = ArrayString::<30>::new();
 
-    let base32 = [0x00u8, 0x01, 0x02].iter().copied().bytes_to_fes().collect::<ArrayVec<u5, 30>>();
+    let data = [0x00u8, 0x01, 0x02];
+    let hrp = Hrp::parse("bech32").expect("failed to parse hrp");
 
-    let hrp = Hrp::parse("bech32").unwrap();
-
-    bech32::encode_to_fmt_anycase(&mut encoded, hrp, &base32, Variant::Bech32).unwrap().unwrap();
+    bech32::encode_to_fmt::<Bech32, ArrayString<30>>(&mut encoded, hrp, &data)
```
apoelstra, line 26: In 4da0a7b9784a7fe92db39a7d645d4bee97999d4d:

Could use `<Bech32, _>` rather than repeating `ArrayString<30>`.

src/primitives/mod.rs
```diff
@@ -26,6 +26,11 @@ pub enum Bech32 {}
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub enum Bech32m {}
 
+/// The codex32 checksum algorithm, defined in [BIP-93].
+/// [BIP-350]: <https://github.com/bitcoin/bips/blob/master/bip-0093.mediawiki>
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum Codex32 {}
```
apoelstra, line 32: In 4da0a7b9784a7fe92db39a7d645d4bee97999d4d:

This can live in a doccomment or something but we probably shouldn't export it from this library.

src/lib.rs
```diff
@@ -30,1209 +106,244 @@ extern crate core;
...
 pub use crate::primitives::hrp::Hrp;
+#[doc(inline)]
 pub use crate::primitives::iter::{ByteIterExt, Fe32IterExt};
-pub use crate::primitives::{Bech32, Bech32m};
+#[doc(inline)]
+pub use crate::primitives::{Bech32, Bech32m, NoChecksum};
 
 mod error;
+/// Re-exports the hrp types from [`primitives::hrp`] to make importing ergonomic for the top level APIs.
+pub mod hrp;
```
clarkmoody, line 128: Love the re-exports to the top-level `hrp` module.

Do the doc comments need to point out that these are re-exports?

src/primitives/mod.rs
```diff
@@ -26,6 +26,11 @@ pub enum Bech32 {}
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub enum Bech32m {}
 
+/// The codex32 checksum algorithm, defined in [BIP-93].
+/// [BIP-350]: <https://github.com/bitcoin/bips/blob/master/bip-0093.mediawiki>
```
clarkmoody, line 30: Should this be `[BIP-93]` here?

src/lib.rs
```diff
@@ -30,1209 +106,244 @@ extern crate core;
...
+/// All the primitive types and functionality used in encoding and decoding.
 pub mod primitives;
+/// API for encoding and decoding segwit addresses.
```
clarkmoody, line 131: In the docs, this comment collides with the existing module-level docs in `segwit.rs`, giving us:

> API for encoding and decoding segwit addresses. Segregated Witness API - enables typical usage for encoding and decoding segwit addresses.

We should remove one of the comments to avoid the redundant sentence.

clarkmoody (commented):
Nice top-level examples in the doc code!

src/lib.rs
apoelstra: There is a stupid compiler warning now if re-exports don't have doccomments. Since they appear in the docs under `Re-exports` with a link to their definition, this isn't really sensible, so we've been putting "placeholder comments" like this in rust-bitcoin.

embedded/no-allocator/src/main.rs
tcharding: Done, thanks.

src/primitives/mod.rs
tcharding: Done.

src/primitives/mod.rs
tcharding: Thanks, odd that `-D rustdoc::broken-intra-doc-links` didn't catch that. I've moved the codex32 stuff to a rustdoc comment and remove the link altogether.

src/lib.rs
tcharding: Ha, I saw this once and didn't fix it. Thanks for taking the time to build the docs and check them!

I've removed the docs from `rust-bech32/src/hrp.rs` and clippy seems to be ok with it, although I'm unsure why. Like @apoelstra said above we saw a new lint trigger for this situation in rust-bitcoin, will see how it goes in CI.

src/lib.rs
tcharding: Commented on this below.

tcharding:
> In [4da0a7b](https://github.com/rust-bitcoin/rust-bech32/commit/4da0a7b9784a7fe92db39a7d645d4bee97999d4d):
> 
> * I'm a bit confused why `decode` and `decode_bech32m`/`decode_bech32`.

Ok, I removed them and added more docs about decoding with the checksum algorithm to the `decode` docs.

> * Could you add an analogous `encode_to_write` / `decode_to_read` pair to `encode_to_fmt`? Then no-alloc users could encode the string either to an `io::Write` or `fmt::Write`. (Annoyingly this will require a std gate, but that's ok, it would still be useful I think.)

Done!

> Otherwise this looks great!

Thanks man, this force push was the result of a pretty wild session, you can wait to review until I've had a chance to re-look at it with fresh eyes to catch my own mistakes if you like.

src/segwit.rs
```diff
@@ -100,6 +102,19 @@ pub fn encode(
...
+///
+/// Reads a string from reader and then decodes it (see [`crate::decode`]).
+#[cfg(feature = "std")]
+#[inline]
+pub fn decode_from_reader<R: std::io::Read>(
+    r: &mut R,
+) -> Result<(Hrp, Fe32, Vec<u8>), DecodeFromReaderError> {
+    let mut s = String::new();
+    r.read_to_string(&mut s)?;
+    Ok(decode(&s)?)
```
apoelstra, line 115: In 68c4dfd2666ff390cc10d14adc7d5302667e3f20:

I'll go ahead and ACK this since it's not API-visible, but the string allocation (and UTF-8 checks and everything that comes with it) are all unnecessary here. You should be able to just use the `bytes()` iterator on the `Read`, adapt it wiht `char::from` to get characters, then use the iterator API.

apoelstra (approved):
ACK 68c4dfd2666ff390cc10d14adc7d5302667e3f20

src/segwit.rs
tcharding: I like it! I'll have go.

tcharding: I had a bit of a play with this, unless I'm missing something:
- The iterator API (you mean `iter.rs`, right?) is for encoding not decoding
- The API in `decode.rs` could be changed to enable decoding from a reader but its non-trivial because all the decoding types (eg, `UncheckedHrpstring`) have lifetimes that expect a string somewhere to reference.

We could read into a buffer because a bech32 string is max 90 chars, then parse it out however we like.

Or am I totally missing something?

apoelstra: No, you may be correct. Several iterations ago I had an idea for decoding through the iterator API and I guess that got dropped.

I think it's fine to just drop this function then. I don't have a clear usecase for it and it doesn't even work with nostd.

tcharding: Removed, both in `lib.rs` and `segwit.rs`

apoelstra (approved):
ACK 9a90d64b6f182d31c80c67d80c831862b772c390

tcharding:
Rebaes only, no other changes.

apoelstra (approved):
ACK 79918f664785ce6b4589842713898db9ea93259b

clarkmoody (approved):
ACK 79918f6

PR #129, opened 2023-09-11
segwit: Inline functions

tcharding:
As we do, inline "small" public functions in the `segwit` module.


apoelstra (approved):
ACK a774bc1784f844bb46d558df08acf9c0d039fbb7

PR #130, opened 2023-09-18
Fix index out of bounds bug

tcharding:
Noob bug here, I accessed an array without first checking that it was non-empty :(

Adds unit test as a separate patch, re-arrange to verify the bug. 

Discovered while fuzzing #128.

### DIFF
--- a/src/primitives/decode.rs
+++ b/src/primitives/decode.rs
@@ -947,4 +947,26 @@ mod tests {
             assert!(CheckedHrpstring::new::<Bech32>(valid).is_ok())
         }
     }
+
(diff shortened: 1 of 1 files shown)


apoelstra:
Nice, win for fuzzing. I guess I should've caught this in review -- though it's tough with big PRs like this, especially with lots of iteration.

apoelstra (approved):
ACK ea644f05a402a5f934c5c652680ab2d92db2ed2e

apoelstra:
Will go ahead and merge this without @clarkmoody since it's small and purely a bugfix.

PR #132, opened 2023-10-04
Audit error types

tcharding:
Audit all error types and ensure the following holds:

- All use `non_exhaustive`
- All derive `Debug, Clone, PartialEq, Eq` (unless `io::Error` is present, in which case only `Debug`)
- All error `From` impls use `inline`

This is similar to https://github.com/rust-bitcoin/rust-bitcoin/pull/2101, probably should have done this one first.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -345,7 +345,8 @@ pub fn encode_upper_to_writer<Ck: Checksum, W: std::io::Write>(
 
 /// An error while decoding an address.
 #[cfg(feature = "alloc")]
-#[derive(Debug)]
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
 pub enum DecodeError {
     /// Parsing failed.
     Parse(UncheckedHrpstringError),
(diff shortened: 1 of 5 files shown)


apoelstra (approved):
ACK 9b23c9c230860d7bc9c95202dea2f6a441833227

dpc:
BTW. You could use semgrep to enforce it. [Example from Fedimint](https://github.com/fedimint/fedimint/blob/master/.config/semgrep.yaml). We've found it quite useful.

tcharding:
Thanks for the link, can `semgrep` enforce things not being in the code? Looks like it just flags things that exist in the code. Most of the error stuff here is things that we want to enforce existing instead of things to avoid.

dpc:
> Thanks for the link, can `semgrep` enforce things not being in the code?

As long as you match around something that should always be there, then you can make an exception for the more specific thing that is correct.

[Here](https://github.com/fedimint/fedimint/blob/cdd2b23679fb8605a1fb904953c98bb3b7a3caa2/.config/semgrep.yaml#L75) we match all logs statements, but then let them pass if they have the `target: xyz` that we require.

@tcharding

tcharding:
Thanks for clarifying, we could definitely use this.

PR #137, opened 2023-10-10
Improve public re-exports

tcharding:
Improve the public exports in two ways:
    
1. Inline re-exports into the docs of the module that re-exports them.
2. Separate public and private use statements

Recently we discussed a way to separate the public and private import statements to make the code more clear and prevent `rustfmt` joining them all together.

Separate public exports using a code block and `#[rustfmt::skip]`. Has the nice advantage of reducing the number of `#[doc(inline)]` attributes also.

1. Modules first, as they are part of the project's structure.
2. Private imports
3. Public re-exports (using `rustfmt::skip` to prevent merge)

Use the format

```rust
mod xyz;
mod abc;

use ...;

pub use {
    ...,
};
```

This patch introduces changes to the rendered HTML docs.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -130,28 +130,6 @@ extern crate alloc;
 #[cfg(any(test, feature = "std"))]
 extern crate core;
 
-#[cfg(all(feature = "alloc", not(feature = "std"), not(test)))]
-use alloc::{string::String, vec::Vec};
-use core::fmt;
-
-use crate::error::write_err;
-#[doc(inline)]
-pub use crate::primitives::checksum::Checksum;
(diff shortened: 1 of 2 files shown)


tcharding:
Based on discussion here: https://github.com/rust-bitcoin/rust-bitcoin/discussions/2088

clarkmoody (approved):
Don't have a strong opinion on this one, but I'm fine with it.

ACK 1d4d3372552da33848bc9b4aa6a08ae046f13990

apoelstra (approved):
ACK 1d4d3372552da33848bc9b4aa6a08ae046f13990

PR #138, opened 2023-10-16
Add buffering during encode

tcharding:
Currently we write char at a time to both writers (`fmt::Write` and `std::io::Write`). We can improve performance by first collecting the ASCII bytes of the encoded bech32 string into a buffer on the stack then writing the buffer as a single call.
    
This is purely an optimization.

The second patch is surprising, at least to me, would love to learn what I'm doing wrong. I thought this would be valid
```rust
let mut buf = [0_u8; Ck::CODE_LENGTH]
```

But the compiler does not like the usage of an associated const.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -168,6 +168,12 @@ pub use {
     crate::primitives::{Bech32, Bech32m, NoChecksum},
 };
 
+// Write to fmt buffer, small during testing to exercise full code path.
+#[cfg(not(test))]
+const BUF_LENGTH: usize = 1024;
+#[cfg(test)]
+const BUF_LENGTH: usize = 10;
+
 /// Decodes a bech32 encoded string.
 ///
 /// If this function succeeds the input string was found to be well formed (hrp, separator, bech32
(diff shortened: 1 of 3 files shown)


tcharding:
Fuzzing for the win, uncovered #140

tcharding:
Needs redoing on top of #142

apoelstra:
I'm not thrilled with the casts from `char` to `u8`. There are a couple approaches we could take:

* Add a `bytes` iterator alongside the `chars` iterator that yields `u8`s directly
* Copy from https://github.com/rust-bitcoin/hex-conservative/pull/41 which calls `as_bytes` and `len` on a `char` to turn it into a byteslice which can be copied.

Alternately I can just ACK this as-is. What do you think?

apoelstra:
ACK a957f7f7aae6177eec96c18dcb30bbe16b01f5e2 other than that.

tcharding:
> I'm not thrilled with the casts from `char` to `u8`. 

Why? `HRP` is checked to be valid ASCII and the fe32s are valid ASCII also so its safe to cast.

apoelstra:
@tcharding because every time I read the code I need to mentally re-check those facts. I agree it's safe, but it's a code smell.

tcharding:
Ah yes, I guess I did it in some place without code comments. I'll see if I can come up with something less smelly.

tcharding:
Added `primitives::encode::ByteIter` and used that to remove the casts in `crate::segwit`.

apoelstra:
Perfect, thanks! I'm a tiny bit disappointed that `ByteIter` goes through `CharIter` rather than directly converting fes to bytes and pulling bytes out of the `Hrp`, but I see that this would be a ton of complexity, and anyway it's just an optimization that's unlikely to be measurable.

apoelstra (approved):
ACK c215c3d596761a4f4bf32be62ffbacd17081fe41

apoelstra:
I wonder if we could stick a `.buffered(buffer_size)` adaptor onto `CharIter` (or whatever our "final" encoding iterator is), which wouldn't be an iterator anymore but would be `fmt::Write` and `io::Write`.

tcharding:
Flagging this, is the comment above actionable before  #168 merges?

apoelstra:
@tcharding no. The more I think about this the less clear it is to me what we actually want to do here, buffering-wise. Let's just do 0.10.0 and then I think we'll be "done" our part (except for the error correction stuff which I continue to have on my todo list) and can ask Kix to come take a look.

PR #141, opened 2023-10-17
Check for changes to the public API

tcharding:
We would like to get to a stage where we can commit to the public API. To help us achieve this add a script that generates the public API and checks it against three committed files, one for each feature set: no features, alloc, std.

The idea is that with this applied any PR that changes the public API should include a final patch that is just the changes to the api/*.txt files, that way reviewers can discuss the changes without even needing to look at the code, quickly giving concept ACK/NACKs. We also run the script in CI to make sure we have not accidentally changed the public API so that we can be confident that don't break semver during releases. The script can also be used to diff between two release versions to get a complete list of API changes, useful for writing release notes and for users upgrading.

There is a development burden involved if we apply this patch.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -140,3 +140,18 @@ jobs:
         run: cargo install cross --locked
       - name: run cross test
         run: cross test --target s390x-unknown-linux-gnu
+
+  API:
+    name: Check for changes to the public API
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+      - name: Checkout Crate
+        uses: actions/checkout@v3
+      - name: Checkout Toolchain
+        uses: dtolnay/rust-toolchain@nightly
+      - name: Install cargo-public-api
+        run: cargo install --locked cargo-public-api
+      - name: Running API checker script
+        run: ./contrib/check-for-api-changes.sh
--- a/README.md
+++ b/README.md
@@ -33,3 +33,10 @@ Alternatively add symlinks in your `.git/hooks` directory to any of the githooks
 
 We use a custom Rust compiler configuration conditional to guard the benchmark code. To run the
 benchmarks use: `RUSTFLAGS='--cfg=bench' cargo +nightly bench`.
+
+
+## API changes
+
+All PRs that change the public API of `rust-bech32` must include a patch to the
+`api/` text files. For PRs that include API changes, add a separate patch to the PR
+that is the diff created by running `contrib/check-for-api-changes.sh`.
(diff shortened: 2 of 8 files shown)


tcharding:
This is the same idea as https://github.com/rust-bitcoin/rust-bitcoin/pull/1880

apoelstra (approved):
ACK 62a188d7c8c345dec6a26b5e4812ab2829195aa5

apoelstra:
Definitely need @clarkmoody to weigh in before we can move forward on this :)

tcharding:
There is a fair bit of context in the PR I link to above @clarkmoody, no rush on this one. Thanks

tcharding:
Rebased and added changelog section so this conforms to `mkchlog` (introduced in #155), please do not merge until after #155.

apoelstra:
cc @clarkmoody can you take a look at this?

tcharding:
I can split this up like https://github.com/rust-bitcoin/rust-bitcoin/pull/1880 if its easy for you guys to review.

tcharding:
I can't work out why this is failing. Its giving a different sort order than what is in the PR (ie., I'm getting a different order locally to what the VM in CI is getting)?

apoelstra:
See discussion on https://github.com/rust-bitcoin/hex-conservative/pull/34 (looks like we didn't do this for rust-secp; surprised we haven't had problems there). I think we want `sort -n` which seems to be more consistent between implementations, and then `--unique` doesn't work properly so we want to pipe into `| uniq`?

Regardless of reasoning, we should cargo-cult `sort -n | uniq` from the hex-conservative crate and see if that fixes things.

tcharding:
OMG finally, this sort command got us there `SORT="sort --numeric-sort --ignore-case --stable --unique"`.

apoelstra (approved):
ACK 099b1f84319d52b2caa74f36e9d14348f78a42bb

apoelstra (changes requested):
Rescinding ACK. This produces nearly-empty API .txt files.

tcharding:
Ugh, bollocks.

tcharding:
I'm going to debug this in `hex-conservative`, sorry for all the noise, I had no idea it would be such a PITA.

tcharding:
@clarkmoody are you ok if we merge this? It adds additional burden to contributors with the stated aim of improving our ability to commit to a stable API.

tcharding:
Gentle bump please @clarkmoody, really don't want to merge this without your consent and I'd love to release a non-beta version of this crate `v0.10.0`.

clarkmoody (approved):
ACK 8241316a4c9d02f3c1e35bfb7227808912652b0b

:shipit:

apoelstra (approved):
ACK 8241316a4c9d02f3c1e35bfb7227808912652b0b

Issue #140, opened 2023-10-17
bech32 string maximum length is not enforced


tcharding:
None of our encoding functions enforce the 90 maximum length specified in BIP-173.

We should be enforcing `hrp.len() + 1 + bytes_len_to_fes_len(data.len()) <= 90`

Found while implementing buffered encoding in: https://github.com/rust-bitcoin/rust-bech32/pull/138

PR #142, opened 2023-10-17
Enforce maximum length

tcharding:
Enforce maximum string length when encoding.

Fix: #140

### DIFF
--- a/src/primitives/segwit.rs
+++ b/src/primitives/segwit.rs
@@ -9,6 +9,14 @@ use core::fmt;
 
 use crate::primitives::gf32::Fe32;
 
+/// The maximum enforced string length of a segwit address.
+///
+/// The maximum length as specified in BIP-173, this is less than the 1023 character code length.
+/// This limit is based on empirical error-correcting properties. See ["Checksum design"] section.
+///
+/// ["Checksum design"]: <https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#user-content-Checksum_design>
+pub const MAX_STRING_LENGTH: usize = 90;
+
 /// The field element representing segwit version 0.
 pub const VERSION_0: Fe32 = Fe32::Q;
 /// The field element representing segwit version 1 (taproot).
--- a/src/segwit.rs
+++ b/src/segwit.rs
@@ -45,6 +45,7 @@ use alloc::{string::String, vec::Vec};
 use core::fmt;
 
 use crate::error::write_err;
+use crate::primitives::decode::SegwitCodeLengthError;
 #[cfg(feature = "alloc")]
 use crate::primitives::decode::{SegwitHrpstring, SegwitHrpstringError};
 use crate::primitives::gf32::Fe32;
@@ -52,7 +53,9 @@ use crate::primitives::hrp::Hrp;
 use crate::primitives::iter::{ByteIterExt, Fe32IterExt};
 #[cfg(feature = "alloc")]
 use crate::primitives::segwit;
-use crate::primitives::segwit::{InvalidWitnessVersionError, WitnessLengthError};
+use crate::primitives::segwit::{
+    InvalidWitnessVersionError, WitnessLengthError, MAX_STRING_LENGTH,
+};
 use crate::primitives::{Bech32, Bech32m};
 
 #[rustfmt::skip]                // Keep public re-exports separate.
@@ -79,7 +82,8 @@ pub fn decode(s: &str) -> Result<(Hrp, Fe32, Vec<u8>), DecodeError> {
 
 /// Encodes a segwit address.
 ///
-/// Does validity checks on the `witness_version` and length checks on the `witness_program`.
+/// Does validity checks on the `witness_version`, length checks on the `witness_program`, and
+/// checks the total encoded string length.
 ///
 /// As specified by [`BIP-350`] we use the [`Bech32m`] checksum algorithm for witness versions 1 and
 /// above, and for witness version 0 we use the original ([`BIP-173`]) [`Bech32`] checksum
@@ -101,12 +105,17 @@ pub fn encode(
     segwit::validate_witness_version(witness_version)?;
     segwit::validate_witness_program_length(witness_program.len(), witness_version)?;
 
+    let _ = encoded_length(hrp, witness_version, witness_program)?;
+
     let mut buf = String::new();
     encode_to_fmt_unchecked(&mut buf, hrp, witness_version, witness_program)?;
     Ok(buf)
 }
 
 /// Encodes a segwit version 0 address.
+///
+/// Does validity checks on the `witness_version`, length checks on the `witness_program`, and
+/// checks the total encoded string length.
 #[cfg(feature = "alloc")]
 #[inline]
 pub fn encode_v0(hrp: Hrp, witness_program: &[u8]) -> Result<String, EncodeError> {
@@ -114,6 +123,9 @@ pub fn encode_v0(hrp: Hrp, witness_program: &[u8]) -> Result<String, EncodeError
 }
 
 /// Encodes a segwit version 1 address.
+///
+/// Does validity checks on the `witness_version`, length checks on the `witness_program`, and
+/// checks the total encoded string length.
 #[cfg(feature = "alloc")]
 #[inline]
 pub fn encode_v1(hrp: Hrp, witness_program: &[u8]) -> Result<String, EncodeError> {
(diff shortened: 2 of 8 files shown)


tcharding:
Patch separation for the win! I had a feeling there was going to be something to this that I was missing :) Thanks, will re-spin.

tcharding:
> The maximum length needs to be a parameter of the checksum. 93 is only for bech32.
> 
> And IMO we should only enforce the maximum length of 93 for segwit ... for the non-segwit API we should use 1023 since that's the actual length of the code.

Why do you write 93 here? In BIP-173 it says 90, or am I missing something?

https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki

apoelstra:
Sorry, I was confusing 93 (which is 1023 / 11, a fact which is relevant when searching for BCH codes) with 90 (an empirically derived limit in BIP 173).

tcharding:
> The maximum length needs to be a parameter of the checksum.

I'm stuck. I think I'm hitting a bip/spec wording issue again.

1. `bech32::decode` decodes a bech32/bech32m checksummed string
2. `bech32::encode::<Ck>()` encodes a string and appends a generic checksum

(Note we do not currently provide a way to decode a generic checksummed string without going into `primitives`.)

Questions

- Is a "bech32 encoded string" a string that conforms exactly to BIP-173 or is there a superset of strings that are also included (ie, without the 90 char limit)?
- Same question said differently: is the result of `bech32::encode::<Bech32m>()` a BIP-173 compliant bech32 string?

In my mind the `lib.rs`/`segwit.rs` separation is because of this line in bip-173: "We first describe the general checksummed base32[[1](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#cite_note-1)] format called Bech32 and then define Segregated Witness addresses using it." From that line the strings in `lib.rs` should enforce the limit. I do note however that in v0.9.0 the length was not enforced and I never understood why, I think this is the "there are users who use the `bech32` crate in ways not covered by bip-173" thing.

tcharding:
Deleted that long message to save anyone having to read it. I found my problem

> MAY exceed the 90-character limit specified in BIP-0173.

ref: https://github.com/lightning/bolts/blob/master/11-payment-encoding.md

src/segwit.rs
```diff
@@ -364,6 +378,8 @@ pub enum EncodeError {
     WitnessVersion(InvalidWitnessVersionError),
     /// Invalid witness length.
     WitnessLength(WitnessLengthError),
+    /// Encoding HRP, witver, and program into a bech32 string exceeds limit of 90 characters.
+    TooLong(EncodedLengthError),
```
clarkmoody, line 382: Why not `EncodedLengh` here?

tcharding: You were reviewing while I coded so this comment is stale now since I removed the error type all togther. FTR the policy we have come up with is to always prefix error types with `Error`. 

Anyways, please see the new version for better handling of the maximum string length.

tcharding:
Converted to draft while I push up the changes in the first three patches.

tcharding:
Rebased. Also I pulled a random column width fix out into its own patch because I just mentioned it to @realeinherjar in another PR/repo :)

apoelstra:
OK, I think this is fine. TBH I think we should just enforce the 1023-character maximum and say "screw LN" because if they're actually using an error-correcting code with no length limit then the resulting code is unable to correct even a single error and cannot detect more than one.

apoelstra (approved):
ACK 65790bf4ce0e8b275e984cf4623dfff510f2a5c7

tcharding:
I can do that, can you point me to some docs so I can (a) understand the 1023 limit and (b) reference it in the code?

tcharding:
A real limit is bettor for #138 too, saves looping.

apoelstra:
I don't know where it's documented. But it's a fact of the bech32 and bech32m checksums that they have length 1023. [sipa's find_bch.py can be used to regenerate the code and it outputs some extra data](https://github.com/sipa/pychar2/blob/main/find_bch.py).

codex32 has length 93; the "long codex32" checksum has length 1023.

I'm not sure about the descriptor checksum, I'll have to check.

tcharding:
> I don't know where it's documented.

I found this at the end of bip-173 (in the "Checksum design" section):

> Even though the chosen code performs reasonably well up to 1023 characters, other designs are preferable for lengths above 89 characters (excluding the separator).

I'll add that as a reference for the numbers.

apoelstra:
We can't unconditionally limit strings to 1023. The constant needs to be a property of the checksum.

tcharding:
> We can't unconditionally limit strings to 1023.

Oh, I thought that was what you said before.

> The constant needs to be a property of the checksum.

I don't see how (abstractly not technically). The `segwit` stuff wants to enforce 90 char limit for `Bech32m` but `lib.rs` does not? And the limit is a bip-173 thing not a BCH code thing, right?

tcharding:
Here is your previous comment:

> And IMO we should only enforce the maximum length of 93 for segwit ... for the non-segwit API we should use 1023 since that's the actual length of the code.

I went searching for what "since that's the actual length of the code" meant, and all I could find was the docs in bip-173 "Checksum design" section. I went through a bunch of wiki pages on codes and couldn't find any mention of an inherent "length". Its just to do with how many errors can be detected with a certain length checksum for a certain amount of input data, right?

apoelstra:
https://en.wikipedia.org/wiki/BCH_code#General_BCH_codes defines length, though pretty obtusely. Here `q` is 32 for every code this library supports, `m` is always 2 (though this only affects error correction so maybe some users could define checksums with other values..). So `q^m - 1` is 1023, and the length of a BCH code, being the order of an element in a multiplicative group of size 1023, must be 1023 or one of its factors.

A more useful definition is probably the one in https://secretcodex32.com/docs/2023-08-23--math.pdf on page 18:
> The length of the code is how long a coded message can be (including the checksum!)
for the code to retain its error-correcting properties.

@tcharding 
> I don't see how (abstractly not technically). The `segwit` stuff wants to enforce 90 char limit for `Bech32m` but `lib.rs` does not? And the limit is a bip-173 thing not a BCH code thing, right?
> 

Yes, the limit is defined by bip 173 based on empirical error-correcting properties. It is not related to the design length of the code and non-segwit users can (probably) get away without caring about it. Though they will get somewhat worse error correcting properties. It is a requirement of segwit but for non-segwit bech32 uses we shouldn't enforce it, because the code does continue to work up to length 1023.

> Also, if its on `Checksum` folk that implement the trait would assume it is checked in `lib.rs` functions?

Yes, if we add a `LENGTH` field to the `Checksum` trait, people would assume that it's checked in lib.rs. And if people want to override this they can make their own copy of `Bech32` where they change the field to `usize::MAX` or whatever.

tcharding:
I had an epic rebase fail, not totally sure I didn't loose stuff ...

tcharding:
Force push is fix a couple of things from yesterday's rebase fail.

tcharding:
This one turned out to be pretty complicated, thanks for your patience reviewing.

src/primitives/segwit.rs
```diff
@@ -9,6 +9,14 @@ use core::fmt;
 
 use crate::primitives::gf32::Fe32;
 
+/// The maximum enforced string length of a segwit address.
+///
+/// The maximum length as specified in BIP-173, this is less than the 1023 character code length.
+/// This limit is based on empirical error-correcting properties, see ["Checksum design"] section.
```
apoelstra, line 15: In 6ca06b1ec89eb797e5cafd84846a7d7c1e2e37dd:

I'd grep for `, see`. Almost all of the instances should `. See` :).

tcharding: Ha! I literally was looking at someone else's code yesterday thinking the opposite. I'm going to go ahead and trust you, your English tends to be better than mine :)

tcharding:
Force push is just the docs thing, no other changes.

tcharding:
Please see #156, I think this PR may be buggy. Perhaps we should just drop the length stuff all together from the `Checksum` trait and `lib.rs`? That would still be an improvement over v0.9 (because it enforces 90 chars) but not a regression because it allows any length strings with `lib.rs`?

apoelstra (approved):
ACK 12d7b385c075d44761704089a9cdcc8ad8fd0c35

apoelstra:
> Perhaps we should just drop the length stuff all together from the `Checksum` trait and `lib.rs`? That would still be an improvement

It would still be an improvement, yes, but it would really be better if you couldn't (easily) use checksums for strings for which the checksums' properties didn't apply.

apoelstra:
In any case I would like to move the checksum length discussion into a separate PR, since this one also contains doccomment improvements and segwit stuff and adds the `CODE_LENGTH` field to the `Checksum` trait, all of which we definitely want to do.

PR #144, opened 2023-10-17
Impl `AsRef<[u8]>` for `Hrp`

Alex6323:
In our project we would like to have direct access to the internal buffer of the `Hrp` struct rather than going through the `ByteIter` first. Thanks.


apoelstra:
I think I would prefer a dedicated `as_bytes() -> &[u8]` method and a dedicated `as_str() -> &str` method, rather than using the `AsRef` trait.

The reason is that I don't think it's obvious that the HRP type should have `[u8]` as its `AsRef` target. I think users could just as easily expect the target to be `str`. So we shouldn't use the trait, we should use explicit method names.

Alex6323:
I'm fine with that :+1:

tcharding:
Can you close this in favour of #145?

Issue #143, opened 2023-10-17
Reference to `Hrp::buf` bytes


thibault-martinez:
Hello, thanks for the recent updates that allowed us to start using your crate.

Is there a specific reason why there is no direct access to the `Hrp::buf` bytes through something like `AsRef<[u8]>` or a method?
Looks like our only solution at the moment is to use `byte_iter` and collect it when all we need is just a reference.
I'm happy doing a PR myself, let me know.

Thanks in advance!

apoelstra:
@thibault-martinez no reason, I think we just overlooked this. I think it's a great idea to return a reference to the `[u8]`. I think we can also return a `&str` accessor which would use [`str::from_utf8`](https://doc.rust-lang.org/std/str/fn.from_utf8.html) internally. (We could also use `from_utf8_unchecked` but I think the performance benefit from skipping the check on a short ASCII string isn't worth the reviewer cost of using unsafe code.)

apoelstra:
So yes, for my part if you submitted a PR then I'd ACK it. I will also let @clarkmoody chime in since this is his crate, in case he's opposed for some reason.

PR #145, opened 2023-10-17
Add Hrp::{as_bytes, as_str}

thibault-martinez:
Fixes https://github.com/rust-bitcoin/rust-bech32/issues/143

### DIFF
--- a/src/primitives/hrp.rs
+++ b/src/primitives/hrp.rs
@@ -14,7 +14,7 @@ use alloc::string::String;
 use core::cmp::Ordering;
 use core::fmt::{self, Write};
 use core::iter::FusedIterator;
-use core::slice;
+use core::{slice, str};
 
 /// Maximum length of the human-readable part, as defined by BIP-173.
 const MAX_HRP_LEN: usize = 83;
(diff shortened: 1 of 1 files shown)


src/primitives/hrp.rs
```diff
@@ -148,6 +148,14 @@ impl Hrp {
     #[inline]
     pub fn to_lowercase(&self) -> String { self.lowercase_char_iter().collect() }
 
+    /// Returns this human-readable part as bytes.
+    #[inline]
+    pub fn as_bytes(&self) -> &[u8] { &self.buf }
+
+    /// Returns this human-readable part as str.
+    #[inline]
+    pub fn as_str(&self) -> Result<&str, str::Utf8Error> { str::from_utf8(&self.buf) }
```
tcharding, line 157: We can use expect here because we only store ASCII bytes, then there is no need to return a `Result`.

apoelstra, line 157: We should unwrap this internally and not return a Result. It is impossible for `str::from_utf8` to fail because we check that our HRPs are valid ASCII when we construct them.

tcharding (commented):
Thanks for the contribution, nice to hear that the crate re-write is useful for you! Both these need to use `self.size` so they don't include all the 0s in the buffer. Ie (`&self.buf[..self.size]`).

You could add these two tests to see the problem
```rust
    #[test]
    fn as_str() {
        let s = "arbitraryhrp";
        let hrp = Hrp::parse_unchecked(s);
        assert_eq!(hrp.as_str(), s);
    }

    #[test]
    fn as_bytes() {
        let s = "arbitraryhrp";
        let hrp = Hrp::parse_unchecked(s);
        assert_eq!(hrp.as_bytes(), s.as_bytes());
    }
```

src/primitives/hrp.rs
tcharding: Ha, we were reviewing at the same time :)

tcharding: You mean `expect` right not unwrap?

```rust
    pub fn as_str(&self) -> &str { str::from_utf8(&self.buf[..self.size]).expect("we only store ASCII bytes") }
```

thibault-martinez:
Done. Sorry I have noticed too late that my colleague Alexander also did a PR.

apoelstra (approved):
ACK fedd38c9753e3c4a943572322e826fe23bc5459c

tcharding (approved):
ACK fedd38c9753e3c4a943572322e826fe23bc5459c

PR #146, opened 2023-10-18
Feature guard write_err import

tcharding:
We only use the `write_err` macro when the "alloc" feature is enabled, the import statement should be feature guarded.

Clears a clippy warning.


apoelstra (approved):
ACK c16b68a92982f02c337c84aa0a6bc47aaca97f95

PR #147, opened 2023-10-18
Rename `EncodeError::Write` variant to `Fmt`

tcharding:
We typically refer to a "fmt" being an object that implements `fmt::Write` e.g, `encode_to_fmt()`.

Currently we are using the variant name `Write` for an error variant that holds a `fmt::Error`, while this is correct in a way it is also confusing because we may one day have an `io::Error` (coming from a `io::Write` writer).

Use `EncodeError::Fmt` instead of `Write`.

### DIFF
--- a/src/segwit.rs
+++ b/src/segwit.rs
@@ -297,7 +297,7 @@ pub enum EncodeError {
     /// Invalid witness length.
     WitnessLength(WitnessLengthError),
     /// Writing to formatter failed.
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK 3e729c31ffeb6e0bb7ee46a463b75de673060a94

apoelstra:
Yeah, good call. Glad we did this before a real release.

PR #150, opened 2023-10-22
Fix checksum error type

tcharding:
We have a couple of checksum error problems;

- Unused checksum related variants in `CharError`
- Unuseful error message (duplicated string "invalid checksum")

Fix both at the same time by removing the unused variants from `CharError` and by renaming the `InvalidChecksum` variant to `InvalidResidue` and improving the `Display` impl appropriately.

### DIFF
--- a/src/primitives/decode.rs
+++ b/src/primitives/decode.rs
@@ -174,7 +174,7 @@ impl<'s> UncheckedHrpstring<'s> {
         }
 
         if self.data.len() < Ck::CHECKSUM_LENGTH {
-            return Err(InvalidChecksumLength);
+            return Err(InvalidLength);
         }
 
         let mut checksum_eng = checksum::Engine::<Ck>::new();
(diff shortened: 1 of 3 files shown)


src/primitives/decode.rs
```diff
@@ -750,19 +739,19 @@ impl std::error::Error for CharError {
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[non_exhaustive]
 pub enum ChecksumError {
-    /// The checksum does not match the rest of the data.
-    InvalidChecksum,
+    /// The checksum residue is not valid for the data.
+    InvalidResidue,
     /// The checksum is not a valid length.
-    InvalidChecksumLength,
+    InvalidLength,
```
apoelstra, line 745: In 8eb7954da33cbb9c5e620a049cff12cb1e1ff94b:

Here and in the display output, we should say "the checksummed string is not a valid length" or "the string is not a valid length for this checksum".

tcharding: I went with "the checksummed string is not a valid length". Thanks.

tcharding:
Hope you weren't still reviewing, I force pushed the string change as suggeseted.

apoelstra (approved):
ACK eeb434f37b87df52e8a9574107b2840fe0a1e5ef

PR #151, opened 2023-10-22
Add `errors` example demo file

tcharding:
Add an `errors` example that creates and prints various error types from the crate. This is useful for two main purposes:
    
- Helps us catch mistakes by exercising error paths.
- Helps us check that error output is useful for debugging.
- Helps us check that output for "alloc" builds is good/useful (no-std)
    
Run with appropriate features to verify the output, e.g.,
```
      `cargo run --example errors --no-default-features --features=alloc`
```

This example is prone to go stale, is it useful? I could not think of any other way to verify that our error output was good (in both std and no-std environments).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,3 +18,7 @@ alloc = []
 
 [target.'cfg(mutate)'.dev-dependencies]
 mutagen = { git = "https://github.com/llogiq/mutagen" }
+
+[[example]]
+name = "errors"
+required-features = ["alloc"]
--- a/examples/errors.rs
+++ b/examples/errors.rs
@@ -1,0 +1,129 @@
+//! Demonstrate output from the various crate errors.
+
+#[cfg(not(feature = "std"))]
+use core::fmt;
+#[cfg(feature = "std")]
+use std::error::Error;
+
+use bech32::{Fe32, Hrp};
+
+fn main() {
+    crate_decode();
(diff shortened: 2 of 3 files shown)


apoelstra:
Hmm, concept ACK 61b6300ad854067c55ef9ce27c3ea3749c74b5c5.

I think in practice this'd be pretty hard to keep in sync with the actual error types. Maybe it should be a collection of unit tests rather than examples?

tcharding:
Its only about the "meaningfulness" of the error messages, I couldn't think of a way to test this except look at it every now and again. Once we are happy I don't think it will matter if some are missed, since devs will tend to copy the error style currently there.

apoelstra:
Ok, makes sense.

After running the example all the error messages look great to me except that we should drop `code=` on the `InvalidChar` message.

tcharding:
Sweet! Thanks for running them.

apoelstra (approved):
ACK a321ee544ba94e18cb2c7678dc2708d5f3482769

apoelstra:
Do you want this to be merged?

tcharding:
Maybe just leave it here and I'll have a go doing it in another repo, then see if it feels like its worth having. I'm not 100% sure it won't just sit and rot never to be touched again.

apoelstra:
I think this has been sitting and rotting long enough :).

PR #153, opened 2023-10-24
Introduce semgrep

tcharding:
This is exploratory, adds a `semgrep` config file with a single rule. The idea is that we could use `semgrep` to enforce coding patterns. The downside is it has to be done with `grep` so it is not general enough to enforce "all error types implement `std::error::Error`" - well at least not trivially AFAICT.

The added rule shows how grep'able things can be enforced.

ref: https://semgrep.dev/orgs/rust_bitcoin/settings/access


tcharding:
CC @dpc for giving me the idea.

contrib/semgrep.yml
```diff
@@ -0,0 +1,10 @@
+rules:
+
+- id: inline-from-impls
+  languages:
+    - rust
+  message: Use inline attribute on `From` implementations.
+  patterns:
+    - pattern-regex: 'From<[A-Za-z]+> for [A-Za-z]+ \{$\n'
```
dpc, line 8: **suggestion:** You don't have to use regex. semgrep supports Rust, but has some limitations around macros handling. This looks like something that should be possible to handle with normal  (non-regex) syntatic matchers.

dpc: @tcharding https://semgrep.dev/playground/s/AdWQ

tcharding: Oh cool, that is way more powerful!

tcharding: I had a play and couldn't work out how to match on attributes (because the `#` is a comment in yaml). Will come back to it. Mentioning unless you know the solution off the top of your head.

dpc: Oh. Yeah. Last time I tried the macros were the tricky one.

dpc: So attribute matching works https://semgrep.dev/playground/s/0z1r , but seems like to in `impl` blocks. :D

I've found some examples in https://github.com/returntocorp/semgrep/blob/develop/tests/patterns/rust/attribute_matching.sgrep

dpc: I feel a bit disgusted by it, but that's the best I was able to get to work: https://semgrep.dev/playground/s/qDGR

Using `focus-metavariable`, `#[$ARGS]`, `$...ARGS`, etc. should be possible to do it better in theory, but each seems to not work with Rust attributes for reasons unknown.

tcharding: ok, thanks man. Will return to this.

tcharding:
The first patch can be removed to see this work in action.

contrib/semgrep.yml
```diff
@@ -0,0 +1,11 @@
+rules:
+
+- id: inline-from-impls
+  patterns:
+    - pattern: |-
+        impl From<$F> for $T { fn from(...) -> $SELF { ... } }
+    - pattern-not-regex: "inline"
```
dpc, line 7: `'\#\[inline\]'` or something like that (to only match `#[inline]`) would be "tighter"

tcharding: Ah yes, nice. Will use, thanks.

tcharding:
FTR this PR is a proof of concept for introducing `semgrep` everywhere. Sorry @clarkmoody this repo is fast becoming the testbed for all the `rust-bitcoin` tooling madness we want to introduce :)

dpc:
> Sorry @clarkmoody this repo is fast becoming the testbed for all the `rust-bitcoin` tooling madness we want to introduce :)

*checks if there's `flake.nix` in the root* :shrug:

tcharding:
You haven't managed to convert me to `nix` yet :)

dpc:
*yet*! :D

PR #158, opened 2023-12-04
Add a returns section to the docs on segwit::decode

tcharding:
We return a 3 element tuple but do not document what it is or the guarntees on the vector length. Do so.


apoelstra (approved):
ACK 6de909e7bd503019371fddd5fe49857e65c25117

PR #159, opened 2023-12-26
Adapt for Qtum

lateminer:
This PR adapts `rust-bech32` crate to work with Qtum bech32 format.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,12 @@
+Rust Bech32 Qtum
+================
+
+Rust implementation of the Bech32 encoding format for Qtum. This repository was forked from [rust-bech32](https://github.com/rust-bitcoin/rust-bech32).
+
+See below for the original `rust-bech32` README.
+
+---
+
 Rust Bech32
 ===========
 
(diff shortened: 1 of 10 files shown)


lateminer:
Sorry, wrong repo.

apoelstra:
It looks like whatever you're doing here is changing the set of standard HRP types and nothing else. You should be able to do this by just depending on this library, creating your own `KnownHrp` type with conversions to `Hrp`, and then using the existing functionality. There should be no need to fork the whole library.

At most you may need to fork/duplicate the segwit module. But the rest is designed so that if you are using a BCH code (even a non-bech32 one!) that uses the bech32 alphabet, everything should Just Work no matter how much stuff you change.

I don't think we'll be able to directly help you much because your project looks like it's technically and morally out of scope for rust-bitcoin, but hopefully this hint will let you avoid a repo fork and possibly allow us to share a bit of development energy.

lateminer:
Sorry again, I accidentally pushed the PR into Bitcoin `rust-bech32` repo instead of Qtum one which I am working on at moment.

I also don't like the idea of forking the whole lib just to change some constants, but considered it as the easiest approach. Thank you very much for giving a direction to look into.

Issue #162, opened 2024-01-03
InvalidWitnessVersion error display should convert fe to u8


apoelstra:
Right now if we have an invalid witness version we print the field element directly, which will print it in the bech32 alphabet. This is confusing because, for example, the number 17 will be printed as "3".

We should convert it to a number by adding `to_u8` before printing.

PR #165, opened 2024-01-07
Improve error display for invalid witness version

tcharding:
We have two errors that cover invalid witness version, for one we print the field element which can be confusing for some values and for the other we omit the invalid version all together - we can do better.

Print the field element as well as the integer value when displaying the two invalid witness version errors.

Fix: #162

### DIFF
--- a/src/primitives/decode.rs
+++ b/src/primitives/decode.rs
@@ -589,6 +589,7 @@ pub enum SegwitHrpstringError {
     Checksum(ChecksumError),
 }
 
+#[rustfmt::skip]
 impl fmt::Display for SegwitHrpstringError {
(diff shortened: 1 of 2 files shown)


apoelstra (approved):
ACK d9bb03ebdf9ac3bb8313a52b7385cf53b527c025

tcharding:
> LGTM. Arguably `fe32` is too jargony. We should maybe say `bech32 character` instead.

yes, good point, will respin.

tcharding:
Added patch up front to skip rustfmt, then used 'bech32 character' as suggested.

apoelstra (approved):
ACK e988f3462f2df36a13ea89a44f5298c0cf6266cf

Issue #160, opened 2024-01-02
UncheckedHrpstring should have a `data` accessor


apoelstra:
This structure is a POD pair of a HRP/data with some utility functions for choosing a checksum and stripping it. To be useful you need to be able to access the underlying data.

Probably same with CheckedHrpstring.

tcharding:
I'm not sure this adds value since the caller can get this by just splitting the input string at the `1` separator and calling `as_bytes()`. Adding this functionality potentially clutters the API (because we the have multiple ways of getting "bytes") and perhaps makes the types harder to understand? Also since these two types are only useful for advanced usage the caller would likely already be quite familiar with the structure of the input bech32 string.

apoelstra:
> I'm not sure this adds value since the caller can get this by just splitting the input string at the `1` separator and calling `as_bytes()`.

..so they need to split the string twice, once when creating the `HrpString` and then once manually? (And then their code needs to be correct wrt the structure of HRP strings)

>Adding this functionality potentially clutters the API (because we the have multiple ways of getting "bytes") and perhaps makes the types harder to understand?

What is the other way to get the underlying bytes of the input string?

apoelstra:
I would also be happy if there were a single method that returned the first byte as a `Fe32` and removed it from the underlying buffer. This is the only missing piece to let the user implement `CheckedHrpstring`/`SegwitHrpstring` themselves.

tcharding:
> removed it from the underlying buffer.

So this piece is missing still, I was not able to fit it into the current abstraction (either in my head or in code :)

apoelstra:
@tcharding look at `CheckedHrpstring::validate_segwit`:

```
        let witness_version = Fe32::from_char(self.data[0].into()).unwrap();
        self.data = &self.data[1..]; // Remove the witness version byte from data.
```

basically I just want those two lines in its own `&mut self` method.

tcharding:
No worries, I can do that.

tcharding:
Done

PR #167, opened 2024-01-09
decode: Add accessors

tcharding:
This is a sexy little PR right here.

Add an `ascii` accessor method to the `UncheckedHrpstring` and `CheckedHrpstring` types. 

- Patch 1 is preparation, renames the `data` field.
- Patch 2 is the meat and potatoes.

Fix: #160

### DIFF
--- a/src/primitives/decode.rs
+++ b/src/primitives/decode.rs
@@ -117,8 +117,8 @@ pub struct UncheckedHrpstring<'s> {
     hrp: Hrp,
     /// This is ASCII byte values of the parsed string, guaranteed to be valid bech32 characters.
     ///
-    /// Contains the checksum if one was present in the parsed string.
-    data: &'s [u8],
+    /// The characters after the separator i.e., the "data part" defined by BIP-173.
+    data_part_ascii: &'s [u8],
     /// The length of the parsed hrpstring.
     hrpstring_length: usize,
 }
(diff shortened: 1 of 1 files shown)


apoelstra:
2be20a198f0680213168e43d9afffb4ec134fb3e looks great! `ascii` is a far better name than `data` for the field.

For the method, can you rename it to `data_part` or `data_part_ascii` (or some variant of this). Just `ascii` implies that we might return the whole string, but we're skipping the HRP and separator.

As a separate request, can you add a commit with method to `UncheckedHrpstring` which returns the witness version and drops it from the `ascii` field? This is ad-hoc and segwit-specific but it's easy to implement. If you don't want to do that, maybe we could add `to_raw_parts` and `from_raw_parts` methods which let the user manually "pop apart" the HrpString and modify it. Or we could do both :).

tcharding:
Ok, I spent  waaay too long on this task. Please see the new ridiculously long function names. Since this is pretty niche usage I think they are warranted - do not ask me how many times I re-wrote this solution.

> As a separate request, can you add a commit with method to UncheckedHrpstring which returns the witness version and drops it from the ascii field? This is ad-hoc and segwit-specific but it's easy to implement. If you don't want to do that, maybe we could add to_raw_parts and from_raw_parts methods which let the user manually "pop apart" the HrpString and modify it. Or we could do both :).

Please see the last patch, I know its not exactly what you specify but is tit enough to serve your needs?

src/primitives/decode.rs
```diff
@@ -250,9 +302,10 @@ impl<'s> UncheckedHrpstring<'s> {
 pub struct CheckedHrpstring<'s> {
     /// The human-readable part, guaranteed to be lowercase ASCII characters.
     hrp: Hrp,
-    /// This is ASCII byte values of the parsed string, guaranteed to be valid bech32 characters,
-    /// with the checksum removed.
-    data: &'s [u8],
+    /// This is ASCII byte values of the parsed string, guaranteed to be valid bech32 characters.
+    ///
+    /// The characters after the '1' separator and the before the checksum.
```
apoelstra, line 307: In 045b50a151bf52c855445e37fd2b359d7ebba5dd:

Typo `the before`

apoelstra (approved):
ACK 2fae3aa5db09806d9e5ab9f604eb1b2bafcbc289

PR #168, opened 2024-01-14
Release tracking PR: `v0.10.0`

tcharding:
In preparation for doing the first "real" (non-beta) release of the new `primitives` module add a changelog entry and bump the version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
-# 0.10.0-beta.2 - 2024-01-17
+# 0.10.0 - 2024-02-22
 
-- Add accessor functions to the `decode` types [167](https://github.com/rust-bitcoin/rust-bech32/pull/167)
+Release the new `primitives` module! This release is a total re-write of the crate - enjoy.
+
+Right before doing this release we added a few API improvements to the `primitives::decode` module:
+
+- Add accessor functions to the `decode` types [#167](https://github.com/rust-bitcoin/rust-bech32/pull/167)
+- Add `CheckedHrpstring::fe32_iter` function [#174](https://github.com/rust-bitcoin/rust-bech32/pull/174)
+- Make a couple of the functions on `CheckedHrpstring` struct public [#173](https://github.com/rust-bitcoin/rust-bech32/pull/173)
 
 # 0.10.0-beta
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bech32"
-version = "0.10.0-beta.2"
+version = "0.10.0"
 authors = ["Clark Moody", "Andrew Poelstra", "Tobin Harding"]
 repository = "https://github.com/rust-bitcoin/rust-bech32"
 documentation = "https://docs.rs/bech32/"


tcharding:
Before this is release can you confirm please that the new accessor methods fully cover your use case in liquid @apoelstra. Using the new API over there was the last step before we were confident to release this, right?

apoelstra:
Yeah, lemme PR to rust-elements to use the new accessors first please.

apoelstra:
@tcharding can we do a `beta2` release first which includes the new stuff?

tcharding:
Sure thing, I couldn't think of a nice next beta version name yesterday - I'll use `beta2` as suggested.

tcharding:
Next beta release done in https://github.com/rust-bitcoin/rust-bech32/pull/169, keeping this here for the `v0.10.0` when it comes - can't be far away.

tcharding:
Note the diff includes change to CHANGELOG removing mention of `v0.10.0-beta.2`, since we yanked it I believe this is correct.

apoelstra:
Ok, sorry, we have to do one more iteration on the API :).

The issue I'm having is that there is a specific sequence to parsing a Segwit string:

* Parse as an `UncheckedHrpstring` to check that the hrpstring is well-formed (can be done)
* First, check (but don't remove) the witness version (can be done with new  `UncheckedHrpstring::witness_version`) (though in retrospect `UncheckedHrpstring::remove_witness_version` I think is useless). Determine the checksum based on the witness version.
* Validate and remove the checksum to get a `CheckedHrpstring` (can be done)
* **Remove the witness version from the `CheckedHrpstring`** (cannot be done)
* Check segwit padding and length (can be done)

So I think what we want to do is move `remove_witness_version` from `UncheckedHrpstring` to `CheckedHrpstring` and we should be good to go.

But we've made a ton of progress :) on my local rust-elements branch we have
```
 5 files changed, 40 insertions(+), 546 deletions(-)
```
which is basically all deleted code that I'd copied and tweaked from this library since the API didn't expose enough for me to use it directly.

tcharding:
>  5 files changed, 40 insertions(+), 546 deletions(-)

That's a win, for sure.

tcharding:
I left the functions on `UncheckedHrpstring` - from an API perspective seems to be no reason to have them on one and not the other.

clarkmoody (approved):
ACK 1b64ed4204e85a8d4907fc92af48e7e1e6eaa79b

tcharding:
ooo yeah, I love waking up in the morning to a fresh release - make my day @apoelstra!

apoelstra (approved):
ACK 1b64ed4204e85a8d4907fc92af48e7e1e6eaa79b let's do it

apoelstra:
Tagged and published.

apoelstra:
Sigh, cargo seems to think that `0.10.0` is the same major rev as `0.10.0-beta` and is breaking downstream crates.

Yanked 0.10.0. Will need to re-release as 0.11.0.

tcharding:
Bother, I wish I came to the same conclusion before I got in a rage about every PR in `rust-bitcoin` being broken.

tcharding:
Like, literally the reason we did the `-beta` releases ....

PR #170, opened 2024-01-16
Use 0.10.0-beta.2 for the upcoming release

tcharding:
We just bumped the version number to `0.10.0-beta2` but that format causes `cargo` to automatically pull the new version. We can use `0.10.1-beta.2` so `cargo` correctly treats it as a new separate version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# 0.10.0-beta2 - 2024-01-16
+# 0.10.0-beta.2 - 2024-01-17
 
 - Add accessor functions to the `decode` types [167](https://github.com/rust-bitcoin/rust-bech32/pull/167)
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bech32"
-version = "0.10.0-beta2"
+version = "0.10.0-beta.2"
 authors = ["Clark Moody", "Andrew Poelstra", "Tobin Harding"]
 repository = "https://github.com/rust-bitcoin/rust-bech32"
 documentation = "https://docs.rs/bech32/"


apoelstra (approved):
ACK 15dfdf0d086c75a59464cf09ee414b5dd283b6cf

apoelstra:
Published and yanked :( same issue.

I think I'll just try to use this version in Elements locally (I think cargo will let me use a yanked version, it'll just whine about it) and if that passes let's just do a real release.

apoelstra:
Ok, cargo won't let me use the yanked version, but I can use my local repo.

apoelstra:
Ok, I have two comments so far:

* The signature is wrong on `data_part_ascii_no_checksum`. It returns a `&[u8]` (which defaults to the lifetime on `&self`) but needs to return `&'s [u8]`. Probably same with the other accessors. Without this fix I get borrowck errors all over the place.
* I would actually like a `Fe32` iterator on `CheckedHrpString`; this is the intermediate iterator used in constructing the return value of `byte_iter` so it'll be easy to implement.

Also it would make my life a tiny bit easier if `validate_padding` and `validate_witness_program_length` were `pub` ... but I don't think we want to do that, it'd make the API too messy and expose too many internals.

tcharding:
I"ll work o the points above. As far as naming, shall we just use `-gamma` like you said before. This is slightly amusing that with our combined number of years writing rust we are having so much difficulty naming a new 'beta' version'.

apoelstra:
I'm doubtful that `gamma` will work when `beta2` didn't, but we might as well try it!

tcharding:
Want to just test the new stuff locally on your end (i.e., after that new stuff above) and the just release the `v0.10.0`?

apoelstra:
Yeah, I think that's best. Next time we'll try starting with `beta.0` etc and see if that works.

tcharding:
> Ok, I have two comments so far:
> 
> * The signature is wrong on `data_part_ascii_no_checksum`. It returns a `&[u8]` (which defaults to the lifetime on `&self`) but needs to return `&'s [u8]`. Probably same with the other accessors. Without this fix I get borrowck errors all over the place.
> * I would actually like a `Fe32` iterator on `CheckedHrpString`; this is the intermediate iterator used in constructing the return value of `byte_iter` so it'll be easy to implement.
> 
> Also it would make my life a tiny bit easier if `validate_padding` and `validate_witness_program_length` were `pub` ... but I don't think we want to do that, it'd make the API too messy and expose too many internals.

All this should be seen to now.

PR #172, opened 2024-01-23
Use correct lifetime

tcharding:
We are returning an reference to the ascii input string, this has a lifetime associated with it - use it.


apoelstra (approved):
ACK 9db843dbddbcdfe84e6e377612586a452cd1baff

PR #175, opened 2024-01-28
Add witness_version functions to `CheckedHrpstring`

tcharding:
As we did for `UncheckedHrpstring` add functions to the `CheckedHrpstring` struct to allow checking the initial byte of the data part and removing it if required.

Add an initial patch that improves the docs on the `UncheckedHrpstring::remove_witness_version` function (this is then copied in patch 2).

### DIFF
--- a/src/primitives/decode.rs
+++ b/src/primitives/decode.rs
@@ -376,6 +376,68 @@ impl<'s> CheckedHrpstring<'s> {
     #[inline]
     pub fn data_part_ascii_no_checksum(&self) -> &'s [u8] { self.ascii }
 
+    /// Attempts to remove the first byte of the data part, treating it as a witness version.
+    ///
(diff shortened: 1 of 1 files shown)


clarkmoody:
Do we need standalone tests for this, or is that captured in the docs examples?

apoelstra:
Let me confirm that breaking the doc examples breaks CI on my end -- but I'm pretty sure they are run in CI, yes.

apoelstra:
Ok, yes, they're run in the real CI but not in my local one :). Glad I checked this. I'll fix the local thing, then remove my failing commit.

apoelstra (approved):
ACK 2d7832ac9aad87c67b0c3be99b9af6966cbc168e tested with rust-elements and it works!

PR #179, opened 2024-03-04
Fixed error message for too long hrp prefix

### LINKED ISSUE #178: Wrong error message
Hi all,

I guess, in [this line](https://github.com/rust-bitcoin/rust-bech32/blob/c2aac494b9932802b99ed648aac34be0822d63e1/src/primitives/hrp.rs#L396)
```
impl fmt::Display for Error {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        use Error::*;

        match *self {
            TooLong(len) => write!(f, "hrp is too long, found {} characters, must be <= 126", len),  // <--- HERE
            Empty => write!(f, "hrp is empty, must have at least 1 character"),
            NonAsciiChar(c) => write!(f, "found non-ASCII character: {}", c),
            InvalidAsciiByte(b) => write!(f, "byte value is not valid US-ASCII: \'{:x}\'", b),
            MixedCase => write!(f, "hrp cannot mix upper and lower case"),
        }
    }
}
```
a previously defined constant should be used, like:
```
impl fmt::Display for Error {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        use Error::*;

        match *self {
            TooLong(len) => write!(f, "hrp is too long, found {} characters, must be <= {}", len, MAX_HRP_LEN),
            Empty => write!(f, "hrp is empty, must have at least 1 character"),
            NonAsciiChar(c) => write!(f, "found non-ASCII character: {}", c),
            InvalidAsciiByte(b) => write!(f, "byte value is not valid US-ASCII: \'{:x}\'", b),
            MixedCase => write!(f, "hrp cannot mix upper and lower case"),
        }
    }
}
```
Currently, such an error is reported when the prefix is too long:
**hrp is too long, found 85 characters, must be <= 126**
which is somewhat confusing ;-).

DariuszDepta:
closes #178

### DIFF
--- a/src/primitives/hrp.rs
+++ b/src/primitives/hrp.rs
@@ -393,7 +393,7 @@ impl fmt::Display for Error {
         use Error::*;
 
         match *self {
-            TooLong(len) => write!(f, "hrp is too long, found {} characters, must be <= 126", len),
+            TooLong(len) => write!(f, "hrp is too long, found {} characters, must be <= {}", len, MAX_HRP_LEN),
             Empty => write!(f, "hrp is empty, must have at least 1 character"),
             NonAsciiChar(c) => write!(f, "found non-ASCII character: {}", c),
             InvalidAsciiByte(b) => write!(f, "byte value is not valid US-ASCII: \'{:x}\'", b),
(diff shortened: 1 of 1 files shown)


apoelstra:
Hi @DariuszDepta, thanks for the contribution! Can you run `cargo fmt` on this?

concept ACK this change.

I'm a little unsure if we want unit tests on the `Display` of our error types. In theory they are part of the public API and they shouldn't change, but on the other hand we may want the flexibility to change them anyway. And if people care about the exact output they should be able to destructure the errors and recreate their own messages.

tcharding:
I also would prefer to see the error variants tested against instead of the error strings, including the strings is a maintenance burden that adds little value IMO (taking into consideration the bit @apoelstra said about technically beig part of the API).

ACK the non-test bit though, thanks for fixing it!

DariuszDepta:
I guess it is ready to merge.

PR #180, opened 2024-03-25
checksum: introduce `PrintImpl` object that can compute various code properties

apoelstra:
This PR does a couple things:

* Bumps the MSRV to 1.56.1 to match rust-bitcoin (and get us const generics which are useful for defining extension fields) as well as cleaning up a couple warnings.
* Introduces the GF1024 field, a simple polynomial type, the Chein root-finding algorithm, and a couple other misc algorithms that will be useful for error correction. (All this is internal and not reflected in the public API.)
* Introduces the GF32768 field which will be needed for error correction of the descriptor checksum.
* To demonstrate this new functionality, adds an object to the API which can produce the `impl Checksum` block for codes which can be corrected using a quadratic or cubic field extension. This will hopefully provide some context for people not super familiar with BCH code about where these magic numbers come from.

We also add unit tests that produce the parameters for codex32 and the descriptor checksum. A later PR will introduce more parameters which will be needed for error correction.


apoelstra:
I updated the API files inline with each commit so it's clear what commits change the API in what ways. Let me know if it'd be better to pull these out into their own commit(s).

apoelstra:
This is ready for review. I extended it a bit to support the descriptor checksum but I'll stop extending it and do future changes in a later PR.

src/primitives/gf32_ext.rs
```diff
@@ -0,0 +1,394 @@
...
+    // (Specifically, the third element of that exp table, 311, expanded into binary,
+    // 01001 10111, and mapped back to Fe32, gives us FH rather than PP. But really,
+    // it doesn't matter, except that it is part of the `Fe1024` API and we cannot
+    // change it once we have published it, since changing it amounts to moving to
+    // a different, though isomorphic, field.)
+    const POLYNOMIAL: Self = Self::new([Fe32::P, Fe32::P]);
+
+    /// The element zeta such that the extension field is defined as GF32\[zeta\].
+    ///
+    /// Alternately, the image of x in the mapping GF32\[x\]/p(x) -> the field
```
tcharding, line 182: nit: The escapes make it pretty noisy in the source code, perhaps we just put backtics around the whole expression?

```rust
    /// Alternately, the image of x in the mapping `GF32[x]/p(x)` -> the field
```

(I built the docs and looked at it, looks ok in HTML.)

tcharding (commented):
That was a pleasure to review, the code is super clean. Code review only, no clue if the math is right :)

src/primitives/gf32_ext.rs
```diff
@@ -0,0 +1,394 @@
...
+use core::{fmt, ops};
+
+use super::field::{ExtensionField, Field};
+use crate::Fe32;
+
+/// An element of the extension field.
+#[derive(Copy, Clone, PartialEq, Eq, Hash)]
+pub struct Fe32Ext<const DEG: usize> {
+    /// The polynomial representation of the element in "little-endian" order;
+    /// that is, the element is the sum inner[i] * EXT_ELEM^i.
```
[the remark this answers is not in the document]
tcharding: Note please, before you change the PR, I make a further contradictory comment below.

src/primitives/field.rs
```diff
@@ -0,0 +1,367 @@
...
+    /// [`ops::Sub`] and [`ops::SubAssign`] traits, which should probably be called
+    /// instead of calling this.
+    fn _sub(&self, other: &Self) -> Self;
+
+    /// Multiplies a value by `self`. This is a helper funcion for implementing the
+    /// [`ops::Mul`] and [`ops::MulAssign`] traits, which should probably be called
+    /// instead of calling this.
+    fn _mul(&self, other: &Self) -> Self;
+
+    /// Divides a value from `self`. This is a helper funcion for implementing the
```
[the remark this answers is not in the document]
apoelstra: Heh, I think the English is a bit sketchy ... but as you say, if you swap the operands there isn't really a good way to phrase it.

src/primitives/field.rs
```diff
@@ -0,0 +1,367 @@
...
+
+    /// The element which is adjoined to the base field to get this field.
+    ///
+    /// In other words, the image of x in the isomorphism from
+    /// `Self::BaseField`\[x\]/`Self::POLYNOMIAL` to `Self`.
+    const EXT_ELEM: Self;
+}
+
+macro_rules! impl_ops_for_fe {
+    (impl$(<$params1:tt>)? for $op:ident$(<$param:tt>)?) => {
```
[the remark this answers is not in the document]
apoelstra: I should probably drop it. It was going to be for if I could implement this generically on `Fe32Ext<DEG>` but then I couldn't figure out how to write a generic-over-`DEG` division algo so I couldn't do it.

src/primitives/field.rs
```diff
@@ -0,0 +1,367 @@
...
+            mask >>= 1;
+        }
+        ret
+    }
+
+    /// The multiplicative order of an element.
+    fn multiplicative_order(&self) -> usize {
+        for &ord in Self::MULTIPLICATIVE_ORDER_FACTORS {
+            if self.powi(ord) == Self::ONE {
+                return ord as usize;
```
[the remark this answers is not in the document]
apoelstra: I think I should make the factors all be `usize`s.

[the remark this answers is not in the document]
apoelstra: But all these casts are OK because in practice none of these values will ever exceed 32 bits (or even 16 bits, probably).

src/primitives/field.rs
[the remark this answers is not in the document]
tcharding: Ok, lets drop it then, we can also simplify the macro call to just plain old `impl_ops_for_fe(Fe32)`

tcharding:
Meant to mention before, I found the API files very useful when reviewing.

apoelstra:
Addressed all comments.

tcharding (approved):
ACK 02282fa7d1ced7521df51a1b9393df44a4af330a

apoelstra:
@clarkmoody what are next steps for this PR?

.github/workflows/rust.yml
```diff
@@ -54,15 +54,15 @@ jobs:
         run: ./contrib/test.sh
 
   MSRV:
-    name: Test - 1.48.0 toolchain
+    name: Test - 1.56.1 toolchain
```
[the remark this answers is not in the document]
apoelstra: Can you go through and resolve all the comments (at least the nits).

tcharding:
I rekon we can merge this, its been open for a while now.

tcharding:
Fair.

tcharding:
Hi @clarkmoody, is there anything we can do to make reviewing this easier for you? I got a message out of band from FROST devs chasing error correction. I realize this is a massive PR and not at all easy to review.

cc @nickfarrow

src/primitives/field.rs
```diff
@@ -0,0 +1,367 @@
...
+    /// The order of the multiplicative group of the field.
+    const MULTIPLICATIVE_ORDER: usize;
+
+    /// All factors of the multiplicative order, in increasing order.
+    ///
+    /// Include both 1 and the number itself. So for example if you have `n` distinct
+    /// prime factors which each appearing once, this array would have size `2^n`.
+    const MULTIPLICATIVE_ORDER_FACTORS: &'static [usize];
+
+    /// Adds a value to `self`. This is a helper funcion for implementing the
```
clarkmoody, line 52: Spelling: `funcion` in a few places

src/primitives/gf32.rs
```diff
@@ -285,58 +256,43 @@ impl AsRef<u8> for Fe32 {
...
+impl super::Field for Fe32 {
+    const ZERO: Self = Fe32::Q;
+    const ONE: Self = Fe32::P;
+    const GENERATOR: Self = Fe32::Z;
+    const MULTIPLICATIVE_ORDER: usize = 31;
+    const MULTIPLICATIVE_ORDER_FACTORS: &'static [usize] = &[1, 31];
+
+    fn _add(&self, other: &Fe32) -> Fe32 { Fe32(self.0 ^ other.0) }
+    // Subtraction is the same as addition in a characteristic-2 field
+    fn _sub(&self, other: &Fe32) -> Fe32 { Fe32(self.0 ^ other.0) }
```
clarkmoody, line 268: Why not call `_add` within the subtraction function to express the equivalence explicitly, rather than duplicating the implementation?

src/primitives/gf32.rs
```diff
@@ -285,58 +256,43 @@ impl AsRef<u8> for Fe32 {
...
+    fn _mul(&self, other: &Fe32) -> Fe32 {
+        if self.0 == 0 || other.0 == 0 {
+            Fe32(0)
+        } else {
+            let log1 = LOG[self.0 as usize];
+            let log2 = LOG[other.0 as usize];
+            Fe32(LOG_INV[((log1 + log2) % 31) as usize])
```
clarkmoody, line 275: The bare `31` here and in `_div` feels like magic. Should these be named constants?

src/primitives/polynomial.rs
```diff
@@ -0,0 +1,281 @@
...
+impl<F: Field> Polynomial<F> {
+    /// The degree of the polynomial.
+    ///
+    /// For constants it will return zero, including for the constant zero.
+    pub fn degree(&self) -> usize {
+        let mut deg = self.inner.len() - 1;
+        while deg > 0 && self.inner[self.inner.len() - 1 - deg] == F::ZERO {
+            deg -= 1;
+        }
+        deg
```
clarkmoody, line 34: What do you think about using iterators here?

```rust
self.inner.iter().rev().take_while(|el| el == Fe::ZERO).count()
```

src/primitives/polynomial.rs
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    /// Given a BCH generator polynomial, find an element alpha that maximizes the
+    /// consecutive range i..j such that `alpha^i `through `alpha^j` are all roots
+    /// of the polynomial.
+    ///
+    /// (Despite the name, the returned element might not actually be a primitive
+    /// element. For a "primitive BCH code" it will be, but in general not. But
+    /// there is no standard name for the element this function returns, and
+    /// "primitive element" is suggestive.)a
```
clarkmoody, line 115: Nit: trailing `a` after the parenthetical

src/primitives/checksum.rs
```diff
@@ -327,4 +495,42 @@ mod tests {
...
+
+        // Ok, exposition over. The actual unit test follows.
+
+        // Run with -- --nocapture to see the output of this. This unit test
+        // does not check the exact output because it is not deterministic,
+        // and cannot check the code semantics because Rust does not have
+        // any sort of `eval`, but you can manually check the output works.
+        let _s = PrintImpl::<Fe1024>::new(
+            "Bech32",
+            &[Fe32::A, Fe32::K, Fe32::_5, Fe32::_4, Fe32::A, Fe32::J],
```
clarkmoody, line 529: Can we pass `&unpacked_poly` here directly, to avoid duplication/confusion?

src/primitives/checksum.rs
```diff
@@ -83,6 +93,161 @@ pub trait Checksum {
...
+        write!(f, "// Code block generated by Checksum::print_impl polynomial ")?;
+        for fe in self.generator {
+            write!(f, "{}", fe)?;
+        }
+        write!(f, " target ")?;
+        for fe in self.target {
+            write!(f, "{}", fe)?;
+        }
+        f.write_str("")?;
+        writeln!(f, "impl Checksum for {} {{", self.name)?;
```
clarkmoody, line 225: We seem to be missing a newline in here. I'm seeing this in the test output:

```rust
// Code block generated by Checksum::print_impl polynomial ak54aj target qqqqqpimpl Checksum for Bech32 {
    type MidstateRepr = u32; // checksum packs into 30 bits
    const CODE_LENGTH: usize = 1023;
    const CHECKSUM_LENGTH: usize = 6;
    const GENERATOR_SH: [u32; 5] = [
        0x3b6a57b2,
        0x26508e6d,
        0x1ea119fa,
        0x3d4233dd,
        0x2a1462b3,
    ];
    const TARGET_RESIDUE: u32 = 1;
}
```

clarkmoody:
This is some real Big Brain stuff! I've pointed out some minor things here and there.

I apologize for going AWOL for so long.

src/primitives/gf32.rs
apoelstra: Sure, I can use `LOG_ENV.len()`. But 31 is the number of nonzero elements in a 32-element field. I think it's ok to use both 31 and 32 as "magic numbers" within this crate.

src/primitives/polynomial.rs
apoelstra: I like it

src/primitives/checksum.rs
apoelstra: I'll try but IIRC there was some type confusion that prevented me from doing so

apoelstra:
Let me open a PR to get CI working then I'll rebase this and address your comments. In the last couple months there has been a bunch of breakage in check-api and even rustc itself.

src/primitives/gf32.rs
apoelstra: Ah, we have the `Self::MULTIPLICATIVE_ORDER` constant available.

src/primitives/polynomial.rs
apoelstra: Actually, after finangling this enough to be correct, it becomes
```rust
self.inner.len() + 1 - self.inner.iter().rev().take_while(|el| **el == F::ZERO).count()
```
which isn't obviously better.

src/primitives/checksum.rs
apoelstra: Oh, I see, you mean to take the `unpacked_poly` computation from the unit test and put it in the doctest, so that the only magic constants that appear clearly came from the BIP. Sure.

apoelstra:
Rebased and addressed all nits.

src/primitives/polynomial.rs
```diff
@@ -0,0 +1,277 @@
...
+    /// coefficient is the one multiplied by x^i.
+    pub fn new(f: Vec<F>) -> Self { Self { inner: f } }
+}
+
+impl<F: Field> Polynomial<F> {
+    /// The degree of the polynomial.
+    ///
+    /// For constants it will return zero, including for the constant zero.
+    pub fn degree(&self) -> usize {
+        self.inner.len() - 1 - self.inner.iter().rev().take_while(|el| **el == F::ZERO).count()
```
clarkmoody, line 30: Maybe an intermediate variable?

```rust
let trailing_zero_terms = self.inner.iter().rev().take_while(|el| **el == F::ZERO).count();
// A degree-N polynomial has N+1 terms
self.inner.len() - trailing_zero_terms - 1
```

Q: Does this go out of bounds when all terms are zeros?

apoelstra: It will panic if there is an empty array, yes, but this should never happen. (The zero polynomial has one term which should be `Fe32(0)`.) I can add a debug_assert to make this explicit.

apoelstra: Pushed to add a debug assertion and named variables.

clarkmoody (approved):
ACK caefc449b77fd6dd7dea7755c1023ec44de26750

:shipit:

tcharding:
Thanks @clarkmoody, way to go.

tcharding:
I'm not going to re-review this, I don't think a shallow re-ack from me adds much value.

PR #183, opened 2024-07-05
fixes for last couple months of CI changes

apoelstra:
Clippy, public-api and cargo.

### DIFF
--- a/src/primitives/decode.rs
+++ b/src/primitives/decode.rs
@@ -720,19 +720,6 @@ impl<'s> ExactSizeIterator for ByteIter<'s> {
     fn len(&self) -> usize { self.iter.len() }
 }
 
-/// An iterator over a parsed HRP string data as field elements.
-pub struct Fe32Iter<'s> {
-    iter: AsciiToFe32Iter<'s>,
-}
-
-impl<'s> Iterator for Fe32Iter<'s> {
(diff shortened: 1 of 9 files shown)


clarkmoody (approved):
ACK ca3f2ef353c1daa49b82e4757785ce2d930403bb

This should reduce some of the noise in #180

src/primitives/decode.rs
```diff
@@ -720,19 +720,6 @@ impl<'s> ExactSizeIterator for ByteIter<'s> {
...
-}
-
-impl<'s> Iterator for Fe32Iter<'s> {
-    type Item = Fe32;
-    #[inline]
-    fn next(&mut self) -> Option<Fe32> { self.iter.next() }
-    #[inline]
-    fn size_hint(&self) -> (usize, Option<usize>) { self.iter.size_hint() }
-}
-
```
tcharding, line 735: Interesting, that is twice this week that code that I have worked on includes types that are not constructable. Seems this should be caught by the linter.

apoelstra: It's a recent addition (only last couple weeks) to the rustc "dead code" lint. It's actually pretty clever because this is a publicly exported type -- there's just no publicly-exported function that returns it.

PR #185, opened 2024-07-16
Add Fe32::ZERO const

optout21:
Add `Fe32::ZERO` const, which is the same as `Fe32::Q`, but with more intuitive name.
Also replace `Fe32::Q` to `Fe32::ZERO` in a few relevant places.

### DIFF
--- a/api/all-features.txt
+++ b/api/all-features.txt
@@ -755,6 +755,7 @@ pub const bech32::primitives::gf32::Fe32::X: bech32::primitives::gf32::Fe32
 pub const bech32::primitives::gf32::Fe32::Y: bech32::primitives::gf32::Fe32
 pub const bech32::primitives::gf32::Fe32::Z: bech32::primitives::gf32::Fe32
 pub const bech32::primitives::gf32::Fe32::ZERO: Self
+pub const bech32::primitives::gf32::Fe32::ZERO: bech32::primitives::gf32::Fe32
 pub const bech32::primitives::gf32::Fe32::_0: bech32::primitives::gf32::Fe32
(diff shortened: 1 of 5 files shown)


src/primitives/decode.rs
```diff
@@ -180,7 +180,7 @@ impl<'s> UncheckedHrpstring<'s> {
     ///
     /// let mut unchecked = UncheckedHrpstring::new(&addr).unwrap();
     /// let witness_version = unchecked.remove_witness_version().unwrap();
-    /// assert_eq!(witness_version, Fe32::Q);
+    /// assert_eq!(witness_version, Fe32::ZERO);
```
apoelstra, line 183: In f2df1282a553bd019e98adaffe9759eb900d0396:

Everywhere it is used as a witness version it should still be `Q` because that is what shows up in the actual encoded address.

apoelstra:
concept ACK, though several of the changes from Q to ZERO I think make the code less clear.

src/primitives/decode.rs
optout21: Done

optout21:
Reverted some of the renames as requested.

apoelstra (approved):
ACK 2d2d8aa7b4e825e9bc5ccf1f0ce3340a5c51fd2e

tcharding:
NACK from me.

How is this bettehanks for the contribution. In my personal opinion patch 2 is a NACK (I'm not a maintainer here though). How is this better than using `Fe32::Q`. Also it begs the question why not have numerics for all 24 characters, and that would be horrible IMO.

clarkmoody:
I also don't see the need for this, unless I'm missing something.

optout21:
Retracting

PR #190, opened 2024-07-31
Add optional dependency on ordering crate

tcharding:
Add an optional dependency on the `ordering` crate and implement `ArbitraryOrd` for `Fe32` if the feature is enabled.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,5 +19,8 @@ alloc = []
 [target.'cfg(mutate)'.dev-dependencies]
 mutagen = { git = "https://github.com/llogiq/mutagen" }
 
+[dependencies]
+ordered = { version = "0.2.2", optional = true}
+
 [lints.rust]
 unexpected_cfgs = { level = "deny", check-cfg = [ 'cfg(bench)', 'cfg(kani)', 'cfg(mutate)' ] }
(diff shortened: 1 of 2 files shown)


tcharding:
Requested by https://github.com/lightningdevkit/rust-lightning/issues/3195

tcharding:
Something is wrong with the check api script, @apoelstra now that we removed it in `rust-bitcoin` what is your thoughts, are we removing it everywhere?

apoelstra:
Yeah, let's remove it everywhere. These constant breakages I think are unacceptable given the way we were trying to use it. I also never got it to work with my local CI so we never really integrated it that well.

apoelstra:
utACK a8ceeb65e91ce06915d556e2a14713884d65675b but the check-API commit should be changed to just remove the script.

tcharding:
>  the check-API commit should be changed to just remove the script.

I don't understand this comment, I've rebased on #192 so that CI passes. If #192 is not right lets discuss it over there.

tcharding:
Rebase only, no code changes.

apoelstra:
FWIW I'm increasingly skeptical of this. It's not clear that anybody needs it and it increases our feature test matrix.

tcharding:
I'll close since you are looking into the `rust-lightning` upgrade to use `bech32 v0.11.0` and that is what this was for.

PR #192, opened 2024-07-31
Remove the check api infrastructure

tcharding:
Turns out the check api stuff did not perform as we had hoped, remove the script and all the output files.

PR #194, opened 2024-07-31
Remove a few casts

### LINKED ISSUE #181: Usage of cast when `From` is available
Using `From` (or `into()`) is easier to read than using casts because it stops devs having to wonder if an data being lost.

We have a few places where we use casts to cast `u8` to `usize`

```rust
            let log1 = LOG[self.0 as usize];
            let log2 = LOG[other.0 as usize];
```

Could be written as:

```rust
            let log1 = LOG[usize::from(self.0)];
            let log2 = LOG[usize::from(other.0)];
```

tcharding:
In an effort to be more readable use `usize::from` instead of a cast. There are a few casts left but this all the ones that can use `From`.

Refactor only, no logic changes.

Close #181

### DIFF
--- a/src/primitives/gf32.rs
+++ b/src/primitives/gf32.rs
@@ -270,8 +270,8 @@ impl super::Field for Fe32 {
         if self.0 == 0 || other.0 == 0 {
             Fe32(0)
         } else {
-            let log1 = LOG[self.0 as usize];
-            let log2 = LOG[other.0 as usize];
+            let log1 = LOG[usize::from(self.0)];
+            let log2 = LOG[usize::from(other.0)];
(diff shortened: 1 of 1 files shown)


clarkmoody (approved):
ACK abac5fd2e60fcafb84ecdd8ba63ff998a58f3d75

`::from` is much better

tcharding:
Force push is rebase only, will still need another one once #192 merges.

tcharding:
Rebase only, no code changes.

apoelstra (approved):
ACK 300efbbd4ff62df73d8aef4cfaaf5e9787a3c241 successfully ran local tests

PR #196, opened 2024-08-06
hrp: add constructor from an arbitrary T: Display

### LINKED ISSUE #195: `Hrp` should have a constructor from a `T: Display`
This would allow easy construction from a rust-lightning `RawHrp`. Currently it [allocates a `String`](https://github.com/lightningdevkit/rust-lightning/blob/9bdce47f0e0516e37c89c09f1975dfc06b5870b1/lightning-invoice/src/ser.rs#L117) which is an unnecessary allocation and inconvenient.

apoelstra:
Fixes #195

### DIFF
--- a/src/primitives/hrp.rs
+++ b/src/primitives/hrp.rs
@@ -119,6 +119,80 @@ impl Hrp {
         Ok(new)
     }
 
+    /// Parses the human-readable part from an object which can be formatted.
+    ///
+    /// The formatted form of the object is subject to all the same rules as [`Self::parse`].
+    /// This method is semantically equivalent to `Hrp::parse(&data.to_string())` but avoids
+    /// allocating an intermediate string.
+    pub fn parse_display<T: core::fmt::Display>(data: T) -> Result<Self, Error> {
+        use Error::*;
+
+        struct ByteFormatter {
+            arr: [u8; MAX_HRP_LEN],
+            index: usize,
+            error: Option<Error>,
+        }
+
+        impl core::fmt::Write for ByteFormatter {
+            fn write_str(&mut self, s: &str) -> fmt::Result {
+                let mut has_lower: bool = false;
+                let mut has_upper: bool = false;
+                for ch in s.chars() {
+                    let b = ch as u8; // cast ok, `b` unused until `ch` is checked to be ASCII
+
+                    // Break after finding an error so that we report the first invalid
+                    // character, not the last.
+                    if !ch.is_ascii() {
+                        self.error = Some(Error::NonAsciiChar(ch));
+                        break;
+                    } else if !(33..=126).contains(&b) {
+                        self.error = Some(InvalidAsciiByte(b));
+                        break;
+                    }
+
+                    if ch.is_ascii_lowercase() {
+                        if has_upper {
+                            self.error = Some(MixedCase);
+                            break;
+                        }
+                        has_lower = true;
+                    } else if ch.is_ascii_uppercase() {
+                        if has_lower {
+                            self.error = Some(MixedCase);
+                            break;
+                        }
+                        has_upper = true;
+                    };
+                }
+
+                // However, an invalid length error will take priority over an
+                // invalid character error.
+                if self.index + s.len() > self.arr.len() {
+                    self.error = Some(Error::TooLong(self.index + s.len()));
+                } else {
+                    // Only do the actual copy if we passed the index check.
+                    self.arr[self.index..self.index + s.len()].copy_from_slice(s.as_bytes());
+                }
+
+                // Unconditionally update self.index so that in the case of a too-long
+                // string, our error return will reflect the full length.
+                self.index += s.len();
+                Ok(())
+            }
+        }
+
+        let mut byte_formatter = ByteFormatter { arr: [0; MAX_HRP_LEN], index: 0, error: None };
+
+        write!(byte_formatter, "{}", data).expect("custom Formatter cannot fail");
+        if byte_formatter.index == 0 {
+            Err(Empty)
+        } else if let Some(err) = byte_formatter.error {
+            Err(err)
+        } else {
+            Ok(Self { buf: byte_formatter.arr, size: byte_formatter.index })
+        }
+    }
+
     /// Parses the human-readable part (see [`Hrp::parse`] for full docs).
     ///
     /// Does not check that `hrp` is valid according to BIP-173 but does check for valid ASCII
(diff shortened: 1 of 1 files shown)


src/primitives/hrp.rs
```diff
@@ -119,6 +119,78 @@ impl Hrp {
...
+                // Unconditionally update self.index so that in the case of a too-long
+                // string, our error return will reflect the full length.
+                self.index += s.len();
+                Ok(())
+            }
+        }
+
+        let mut byte_formatter = ByteFormatter { arr: [0; MAX_HRP_LEN], index: 0, error: None };
+
+        write!(byte_formatter, "{}", data).expect("custom Formatter cannot fail");
```
tcharding, line 184: Doesn't this macro call have to allocate a `String` because `data` is not know until runtime?

My understanding is that `write!` calls `format_args!` which creates an [Arguments](https://doc.rust-lang.org/std/fmt/struct.Arguments.html) type which has an `as_str` method but only if there are no runtime args.

In other words isn't this the same performance as:


```rust
    pub fn parse_display<T: core::fmt::Display>(data: T) -> Result<Self, Error> {
        let s = data.to_string();
        Self::parse(&s)
    }
```

apoelstra: This code works without any allocator. It's not allocating a string. It would also be insane for the `fmt::Write` API to work this way, where every single call to `write_str` involved a gratuitous allocation.

apoelstra: Or indeed, for every call to `write!` anywhere to involve gratuitous allocations.

If you think this could be true we'd need to change our `fmt::Display` implementations throughout every one of our crates to avoid this macro.

apoelstra: It took me a little while to trace through the code but the [core of `write!` is the `write` function](https://doc.rust-lang.org/1.80.0/src/core/fmt/mod.rs.html#1149-1197) which iterates over the "pieces" of an `Arguments` and does not allocate.

tcharding: Cool, TIL. Thanks for the link (I was unconvinced before it, you saved me trawling through the source :)

src/primitives/hrp.rs
```diff
@@ -119,6 +119,78 @@ impl Hrp {
...
+        }
+
+        impl core::fmt::Write for ByteFormatter {
+            fn write_str(&mut self, s: &str) -> fmt::Result {
+                let mut has_lower: bool = false;
+                let mut has_upper: bool = false;
+                for b in s.bytes() {
+                    // Continue after finding an error so that we report the first invalid
+                    // character, not the last.
+                    if !b.is_ascii() {
```
tcharding, line 143: There are a few problems here, I uncovered them by writing a test to check the code comment (which is incorrect, we return the _last_ error), I think we want `break` instead of `continue`.

We need to iterate over `chars` and check they are ascii instead of using `bytes` - we do this in `parse` above.

tcharding: Instead of commenting each line I just wrote some tests that I think should pass

```rust

    #[test]
    fn parse_non_ascii() {
        assert_eq!(Hrp::parse("❤").unwrap_err(), Error::NonAsciiChar('❤'));
    }

    #[test]
    fn parse_display_non_ascii() {
        assert_eq!(Hrp::parse("❤").unwrap_err(), Error::NonAsciiChar('❤'));
    }

    #[test]
    fn parse_display_returns_first_error() {
        assert_eq!(Hrp::parse_display("❤ ").unwrap_err(), Error::NonAsciiChar('❤'));
    }

    // This test shows that the error does not contain heart.
    #[test]
    fn parse_display_iterates_chars() {
        // FIXME: Once test above passes the error char will need changing.
        assert_eq!(Hrp::parse_display(" ❤").unwrap_err(), Error::NonAsciiChar('❤'));
    }
```

apoelstra: Thanks! I will take these tests and make them pass.

apoelstra:
Should be good now.

src/primitives/hrp.rs
```diff
@@ -119,6 +119,80 @@ impl Hrp {
...
+                // invalid character error.
+                if self.index + s.len() > self.arr.len() {
+                    self.error = Some(Error::TooLong(self.index + s.len()));
+                } else {
+                    // Only do the actual copy if we passed the index check.
+                    self.arr[self.index..self.index + s.len()].copy_from_slice(s.as_bytes());
+                }
+
+                // Unconditionally update self.index so that in the case of a too-long
+                // string, our error return will reflect the full length.
```
tcharding, line 178: nit: This comment is stale now I believe.

tcharding (approved):
ACK 2fdc674c41e5a6c5aa6eb8391ade1e4138cdfaa1

src/primitives/hrp.rs
tcharding: We could just check the length at the start and early return.

apoelstra: I don't see why this is stale. The line is called every time the `write_str` function is called, for the reason that the comment describes.

We could do a length check at the start and early return but I don't see what that would accomplish other than increasing the cyclotomic complexity of the function.

tcharding: ooo, I see. Don't mind me.

apoelstra:
I think we may want to backport this to 0.11, since when we release 0.12, rust-bitcoin 0.32 won't be able to use it (thanks to https://github.com/rust-bitcoin/rust-bitcoin/issues/3043).

Same with https://github.com/rust-bitcoin/rust-bech32/issues/197, so let's wait until that's done as well.

src/primitives/hrp.rs
```diff
@@ -538,4 +614,45 @@ mod tests {
...
+    }
+
+    #[test]
+    fn parse_non_ascii() {
+        assert_eq!(Hrp::parse("❤").unwrap_err(), Error::NonAsciiChar('❤'));
+    }
+
+    #[test]
+    fn parse_display_non_ascii() {
+        assert_eq!(Hrp::parse("❤").unwrap_err(), Error::NonAsciiChar('❤'));
```
clarkmoody, line 644: Should this call `parse_display` here?

apoelstra: Oops, yes! Fixed.

tcharding: A bug in the unit tests, what are the odds ;)

tcharding: Oh actually, if this is based on the unit test code I posted above I intentionally included one that parsed using `Hrp::parse`.

clarkmoody (approved):
ACK 11d6cb4dd626d1edd01060e15ac4b7c14a2d30fd

src/primitives/hrp.rs
tcharding: Sweet, cheers. Its hard to get good help :)

PR #200, opened 2024-09-14
A few cleanup commits

apoelstra:
This extracts 3 (hopefully) simple and non-controversial commits from my error-correction branch.

The first is just a clippy fix.

The second cleans up and simplifies the fuzztests (it does *not* bring over the new fuzztesting scripts from rust-bitcoin, use the bitcoin-maintainer-tools repo, introduce libfuzzer, etc.; it just does some basic cleanups to make things more useful).

And the third adds a few new bounds to the `Field` trait, specifically `Neg<Output = Self>` and `Sum`. (This is a breaking change.)

### DIFF
--- a/src/primitives/field.rs
+++ b/src/primitives/field.rs
@@ -2,7 +2,7 @@
 
 //! Generic Field Traits
 
-use core::{fmt, hash, ops};
+use core::{fmt, hash, iter, ops};
 
 /// A generic field.
 pub trait Field:
@@ -13,6 +13,8 @@ pub trait Field:
     + hash::Hash
     + fmt::Debug
     + fmt::Display
+    + iter::Sum
+    + for<'a> iter::Sum<&'a Self>
     + ops::Add<Self, Output = Self>
     + ops::Sub<Self, Output = Self>
     + ops::AddAssign
@@ -29,7 +31,7 @@ pub trait Field:
     + for<'a> ops::MulAssign<&'a Self>
     + for<'a> ops::Div<&'a Self, Output = Self>
     + for<'a> ops::DivAssign<&'a Self>
-    + ops::Neg
+    + ops::Neg<Output = Self>
 {
     /// The zero constant of the field.
     const ZERO: Self;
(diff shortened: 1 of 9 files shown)


apoelstra (commented):
Successfully ran local tests on b2252b7e5f68fbd49b1dbe6026baf447278f61cf.

src/primitives/field.rs
```diff
@@ -362,6 +364,19 @@ macro_rules! impl_ops_for_fe {
...
+            fn sum<I: Iterator<Item = Self>>(iter: I) -> Self {
+                iter.fold(crate::primitives::Field::ZERO, |i, acc| i + acc)
+            }
+        }
+
+        impl<'s> core::iter::Sum<&'s Self> for $op {
+            fn sum<I: Iterator<Item = &'s Self>>(iter: I) -> Self {
+                iter.fold(crate::primitives::Field::ZERO, |i, acc| i + acc)
+            }
+        }
```
tcharding, line 379: Is there any technical reason you chose to not add a `_sum` trait method to the `Field` trait and call it here like the other code does? If not I rekon we should keep it uniform.

tcharding:
Everything else looks good, I even checked `fuzz/generate-files.sh` for possibly the first time ever, to find it doesn't exist :)

src/primitives/field.rs
apoelstra: The reason the `_add` etc methods exist is so that I could implement the 4 different `Add` traits with a single macro call.

There is only one `Sum` trait to implement and it can be implemented in terms of `_add`. I don't see what the value of having a `_sum` method would be. Would I insist that implementors implement it?

apoelstra: If anything I should delete the `_add` methods and improve the macro. I may do this in a followup.

tcharding (approved):
ACK b2252b7e5f68fbd49b1dbe6026baf447278f61cf

src/primitives/field.rs
tcharding: No sweat, cheers.

tcharding: I read the code a bit more, I thought the `_foo` methods were just to make `impl_ops_for_fe` less duplicated, my mistake.

tcharding: Thanks

clarkmoody (approved):
Test suite running on b2252b7e5f68fbd49b1dbe6026baf447278f61cf

Fuzzer seems to be working in CI, but I haven't gotten it working locally.

Changes look reasonable. 

ACK b2252b7e5f68fbd49b1dbe6026baf447278f61cf

apoelstra:
@clarkmoody to get the fuzzer working locally you may need to install a different version of `honggfuzz` and/or run `cargo update -p honggfuzz` in the `fuzz/` directory. My guess is that you are seeing version mismatch errors?

PR #201, opened 2024-09-18
Alternative fes_to_bytes_zeropad with no dropping of last bits

### LINKED ISSUE #198: Option for Fe32->bytes conversion (fes_to_bytes) to not drop last partial values
This is a minor feature request:
The Fe32->bytes conversion -- `fes_to_bytes` in `primitives/iter.rs` -- drops the last incomplete byte:
```
If the total number of bits is not a multiple of 8, any trailing bits are simply dropped
```

In some use cases padding with 0 bits is needed (rather than dropping bits). Currently this can be achieved by appending the input with 0, 1, or 2 `0` values (depending the input length), before doing the conversion.
It would be nice if this is provided as a possibility in conversion (e.g. by a different method `fes_to_bytes_pad` or an parameter flag).

An example with pre-padding is here: https://github.com/lightningdevkit/rust-lightning/blob/main/lightning-invoice/src/lib.rs#L963

If the idea is accepted/clarified, I can prepare a PR.

optout21:
Fixes #198 , as discussed there

### DIFF
--- a/src/primitives/iter.rs
+++ b/src/primitives/iter.rs
@@ -56,7 +56,26 @@ pub trait Fe32IterExt: Sized + Iterator<Item = Fe32> {
     /// are simply dropped.
     #[inline]
     fn fes_to_bytes(mut self) -> FesToBytes<Self> {
-        FesToBytes { last_fe: self.next(), bit_offset: 0, iter: self }
+        FesToBytes {
+            last_fe: self.next(),
+            bit_offset: 0,
+            iter: self,
+            output_incomplete_bits_zeropad: false,
+        }
+    }
+
+    /// Adapts the `Fe32` iterator to output bytes instead.
+    ///
+    /// If the total number of bits is not a multiple of 8, trailing bits
+    /// are padded with the needed amount of zeroes and converted.
+    #[inline]
+    fn fes_to_bytes_zeropad(mut self) -> FesToBytes<Self> {
+        FesToBytes {
+            last_fe: self.next(),
+            bit_offset: 0,
+            iter: self,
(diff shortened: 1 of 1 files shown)


apoelstra:
In 15f6d0ed1fc29a93cd23505a8d629365d01a5bf2:

The change to the doccomment starting with "If the total number of bits" is wrong. (Though I agree this sentence is really long, and if you can break it up somehow, you should.)

Otherwise ack.

CI failure is unrelated and is fixed by #200 (which I will merge shortly, so you will just need to rebase).

optout21:
> The change to the doccomment starting with "If the total number of bits" is wrong. (Though I agree this sentence is really long, and if you can break it up somehow, you should.)

I changed the documentation, made the explanation more detailed:

```rust
    /// Similar to `fes_to_bytes`, but in this variant trailing bits are kept.
    ///
    /// If the last bits do not fill up the last output byte completely, they
    /// are not dropped, but kept, and the last byte is padded with zero bits.
    /// In effect this is the same as if the input was padded with zero values
    /// (1 or 2), such that there are enough zero bits to fill the last byte.
    /// The output is either the same as of `fes_to_bytes`, or has 1 extra byte.
```

apoelstra:
New text looks great! Nah, I wouldn't bother squashing.

apoelstra (approved):
ACK 06762e0df49676f2e0c21930ec8ca79863a10bb2 successfully ran local tests

apoelstra:
Will let @clarkmoody review before merging since this is an API change.

clarkmoody (approved):
ACK 06762e0df49676f2e0c21930ec8ca79863a10bb2

PR #202, opened 2024-09-18
More error correction preparation (upgrades to `Field`, `Polynomial`, `InvalidResidueError` and others)

apoelstra:
This PR has a fairly large diff but the commits should be self-contained and much of the diff consists of comments. In particular this introduces a new type `FieldVec` to back `Polynomial`, which strives to provide a sensible API even in a no-alloc case:

* With an allocator, arbitrarily-large vectors (and therefore arbitrarily large checksums) are supported.
* Without an allocator, the `FieldVec` type panics for large checksums, but since it is only (planned to be) used in error correction, it should be easy to detect this up front and provide a sensible error to the user.

The threshold for "large" is currently set at "bigger than bech32", but it is easy to change this and even easy to add a feature-gate.

I believe this completes all the changes to the traits needed for error correction, and the next PR can just implement correction.


apoelstra:
CI failed because I had a doccomment referring to a field of `Checksum` that didn't yet exist. Backported the commit that introduced it.

But now this PR is way too big :). Gonna close it and split it.

PR #203, opened 2024-09-18
Extend `Field` and `Checksum` for error correction

apoelstra:
The first couple commits of this PR split the `Field` trait into two -- a general `Field` trait which has the algebraic functionality of a field, and a sealed `Bech32Field` trait which has the functionality that is specific to this crate.

The rest add new functionality to the traits which describe extra properties of our checksums needed for error correction.

The next PRs will:

* Introduce a `FieldVec` type to back `Polynomial` to make it work in a no-alloc setting (at least for small checksums, for a configurable notion of "small")
* Implement error correction :)

### DIFF
--- a/src/primitives/field.rs
+++ b/src/primitives/field.rs
@@ -2,6 +2,7 @@
 
 //! Generic Field Traits
 
+use core::iter::{Skip, Take};
 use core::{fmt, hash, iter, ops};
 
 /// A generic field.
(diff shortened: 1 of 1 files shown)


apoelstra (commented):
Successfully ran local tests on f47ec0b9c46c7091e7b9daa192c49e22a28bce6d.

src/primitives/field.rs
```diff
@@ -141,169 +118,205 @@ pub trait ExtensionField: Field + From<Self::BaseField> {
     const EXT_ELEM: Self;
 }
 
+mod private {
+    /// Sealing trait.
+    pub trait Sealed {}
```
clarkmoody, line 139: This is clever 👌

clarkmoody (approved):
ACK f47ec0b9c46c7091e7b9daa192c49e22a28bce6d

Tests running locally. The math is above my pay-grade.

PR #204, opened 2024-09-19
Introduce `FieldVec` and improve `Polynomial` and `InvalidResidueError`

apoelstra:
This introduces a new internal type `FieldVec` which is basically a backing array for checksum residues with some fiddly logic to work in a noalloc context.

Unlike ArrayVec or other similar types, this one's semantics are "an unbounded vector if you have an allocator, a bounded vector otherwise". If you go outside of the bounds without an allocator the code will panic, but our use of it ensures that this is never exposed to a user.

It also assumes that it's holding a `Default` type which lets us avoid unsafety (and dismisses the potential performance cost because this array is never expected to have more than 10 or 20 elements, even when it is "unbounded").

Using this backing, it improves `Polynomial` and removes the alloc bound from it; using this, it puts `Polynomial` into a new `InvalidResidueError` type, which will allow users to extract the necessary information from a checksum error to run the error correction algorithm. (The resulting change to an error type makes this an API-breaking change, though I don't think that anything else here is API-breaking.)

The next PR will actually implement correction :).


src/primitives/polynomial.rs
```diff
@@ -132,75 +135,205 @@ impl<F: Field> Polynomial<F> {
...
+}
+
+impl<F: Field> ops::Add<Polynomial<F>> for Polynomial<F> {
+    type Output = Polynomial<F>;
+    fn add(self, other: Polynomial<F>) -> Polynomial<F> { self + &other }
+}
+
+impl<F: Field> ops::Sub<&Polynomial<F>> for Polynomial<F> {
+    type Output = Polynomial<F>;
+    fn sub(self, other: &Polynomial<F>) -> Polynomial<F> { self + other }
```
tcharding, line 272: This is finite field math so add and sub are the same, right? Perhaps add a comment for less cryptographically trained devs.

tcharding (commented):
My review does not cover the math.

src/primitives/polynomial.rs
```diff
@@ -44,63 +67,43 @@ impl<F: Field> Polynomial<F> {
...
+    /// An iterator over the roots of the polynomial.
     ///
-    /// Does not consider multiplicity; it assumes there are no
-    /// repeated roots. (FIXME we really ought to do so, and
-    /// definitely should before exposing this function in the
-    /// public API.)
+    /// Takes a base element `base`. The roots of the residue will be yielded as
+    /// nonnegative integers between 0 and 1 less than the order of the base,
+    /// inclusive. If `base` is a primitive element of the extension field, then
+    /// all distinct roots (in the extension field) will be found.
```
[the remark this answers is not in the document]
apoelstra: Yes.

src/primitives/polynomial.rs
```diff
@@ -132,75 +135,205 @@ impl<F: Field> Polynomial<F> {
...
-                if r1 == r2 {
+
+        let mut max_length = 0; // length of the max-length geometric series
+        let mut max_start = 0; // i such that the max-length series starts with gen^i
+        let mut max_ratio = 0; // i such that the ratio between terms is gen^i
+
+        for i in 0..roots.len() {
+            for j in 0..roots.len() {
+                if i == j {
                     continue;
```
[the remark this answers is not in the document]
apoelstra: I meant to continue the inner loop.

src/lib.rs
```diff
@@ -134,6 +134,8 @@
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
 // Coding conventions
 #![deny(missing_docs)]
+#![allow(clippy::suspicious_arithmetic_impl)] // this lint is literally always wrong
```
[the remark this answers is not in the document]
apoelstra: No, it's wrong even when doing other kinds of arithmetic. This lint has fired in multiple projects I have worked on and it is literally always wrong.

One time I actually did have a bug where I was using the wrong operator in an operator override impl, and the lint **didn't** fire because I hit some sort of obscure false negative. It's pathetic.

Anyway conceptually the lint is confused. If I was implementing `+` only in terms of other `+` operators I'd hardly need a new type, would I? Or is the argument that Rust has such an anemic set of operator that `+` is used for everything and I shouldn't ever be using non-`+`? That would track but it's not what the lint says.

src/primitives/polynomial.rs
apoelstra: In this case I probably should actually pass through to the sub impl for `F` (which is the same as the add impl in all cases, because I have a binary field, but this stuff should all still work even with non-binary fields).

src/primitives/polynomial.rs
```diff
@@ -2,27 +2,37 @@
...
+    /// Determines whether the residue is representable, given the current
+    /// compilation context.
+    pub fn has_data(&self) -> bool { self.inner.has_data() }
+
+    /// Panics if [`Self::has_data`] is false, with an informative panic message.
+    pub fn assert_has_data(&self) { self.inner.assert_has_data() }
+
+    /// Panics if [`Self::has_data`] is false, with an informative panic message.
+    /// Provide access to the underlying [`FieldVec`].
+    pub fn into_inner(self) -> FieldVec<F> { self.inner }
```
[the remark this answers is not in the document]
apoelstra: No, I think the comment is just out of date. I moved all the panics into functions on `FieldVec` so there's no need for anything in `Polynomial` to directly panic.

src/primitives/polynomial.rs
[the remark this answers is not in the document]
tcharding: Nice, just checking. Thanks.

src/lib.rs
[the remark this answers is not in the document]
tcharding: Warning, I being a massive knob right now.

From the name I assumed it was to catch impls on simple wrapper types where one did a cut'n'pasta and used `+` in `ops::Sub` e.g. the lint fires here as expected 

```rust
use std::ops;

fn main() {
    let one = Foo(1);
    let two = Foo(2);
    let three = Foo(3);

    assert!(three - two == one)
}

struct Foo(u64);

impl ops::Add for Foo {
    type Output = Self;
    fn add(self, rhs: Self) -> Self::Output { Foo(self.0 + rhs.0) }
}

impl ops::Sub for Foo {
    type Output = Self;
    fn sub(self, rhs: Self) -> Self::Output { Foo(self.0 + rhs.0) }
}
```

```bash
% cargo clippy
    Checking play v0.1.0 (/home/tobin/build/play)
warning: suspicious use of `+` in `Sub` impl
  --> src/main.rs:21:58
   |
21 |     fn sub(self, rhs: Self) -> Self::Output { Foo(self.0 + rhs.0) }
   |                                                          ^
```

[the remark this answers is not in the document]
tcharding: > // this lint is literally always wrong

So not _always_ :)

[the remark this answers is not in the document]
apoelstra: Fine. I will change the text to say "this lint is conceptually confused to the point where *at best* it could be saying "double-check this operator impl" on a subset of impls that are more likely to be wrong than the set of all operator impls. It is plausible that, when taken over the entire corpus of Rust code, this could be true. But when taken over the corpus of code written by rust-bitcoin maintainers, it fails even this condition. Instead it fires on code which is correct and never fires on code that is incorrect."

Maybe I should make a new module just for this comment?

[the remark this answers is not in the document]
apoelstra: I am being sarcastic above, of course.

But I would be willing to remove the word "literally" because in the case below (where I was using `+` for `Polynomial::sub` instead of passing through to the underlying field subtraction) it would have been accidentally correct. But the code in question was not copy-pasted, was not a "dumb wrapper type", and it was surrounded by other impls that were correct but also would have triggered the lint.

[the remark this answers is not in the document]
apoelstra: Well, let me sleep on it. I hate this lint so much. It's just rubbing sand in the face of everybody who wanted a sane set of operators for this language. It doesn't complain when people use `+` for concatenation but it does when people use it for pretty-much anything that somebody might describe as "addition".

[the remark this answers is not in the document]
tcharding: \me chuckles

I'm just that guy trying to save us from future hate from the whole entire Rust community :)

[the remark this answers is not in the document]
apoelstra: The existing text is already far more polite than is warranted.

I can remove the word "literally" if you believe it is technically incorrect (though I note your only evidence is hypothetical code that you assume the lint is for). But that is as far as I am willing to go in the direction of weakening my claim.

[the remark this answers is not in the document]
tcharding: lol, love it. Lets just leave it as is :)

apoelstra:
Clicked "resolve" on the conversations we agreed were not actionable.

Force-pushed to address the other comments.

apoelstra (commented):
Successfully ran local tests on 7a28bacacc74e3774a0b119dd83344be41c792e7.

apoelstra:
Second force-push was to fix a bug converting residues to polynomials which I noticed when working on the next PR.

apoelstra (commented):
Successfully ran local tests on 8ee88fcb56e4e8e01586020c70a2f15b1b5cce44.

apoelstra (commented):
Successfully ran local tests on 8ee88fcb56e4e8e01586020c70a2f15b1b5cce44.

apoelstra:
Ok, I have working error correction code now based on this PR :). I will stop iterating on it until somebody reviews it.

apoelstra (commented):
Successfully ran local tests on 8ee88fcb56e4e8e01586020c70a2f15b1b5cce44.

tcharding:
Is a further non-math review useful by me, I can do it if so. I don't want to bother the PR with petty concerns if not.

apoelstra:
I think it'd be valuable, yeah, if you're willing to take the time.

tcharding:
I'll make a coffee :)

src/primitives/fieldvec.rs
```diff
@@ -0,0 +1,380 @@
...
+
+use super::Field;
+use crate::primitives::correction::NO_ALLOC_MAX_LENGTH;
+
+/// A vector of field elements.
+///
+/// Parameterized by the field type `F` which can be anything, but for most methods
+/// to be enabled needs `Default` and `Clone`. (Both are implied by `Field`.)
+#[derive(PartialEq, Eq, Clone, Debug, Hash)]
+pub struct FieldVec<F> {
```
tcharding, line 72: Why go to trouble of mentioning that this should not be exposed in the public API and make this type `pub`? Code seems to build with `pub(crate)`. Is it because of upcoming work?

src/primitives/fieldvec.rs
```diff
@@ -0,0 +1,380 @@
...
+}
+
+impl<F: Field> FieldVec<F> {
+    /// Constructor from the powers of an element, from 0 upward.
+    #[inline]
+    pub fn from_powers(elem: F, n: usize) -> Self {
+        iter::successors(Some(F::ONE), |gen| Some(elem.clone() * gen)).take(n + 1).collect()
+    }
+
+    /// Multiply the elements of two vectors together, pointwise.
```
tcharding, line 146: Needs panics docs.

src/primitives/fieldvec.rs
```diff
@@ -0,0 +1,380 @@
...
+    #[inline]
+    pub fn mul_assign_pointwise(&mut self, other: &Self) {
+        assert_eq!(self.len, other.len, "cannot add vectors of different lengths");
+        for (i, fe) in self.iter_mut().enumerate() {
+            *fe *= &other[i];
+        }
+    }
+
+    /// Multiply the elements of two vectors together, pointwise.
```
tcharding, line 155: And panic docs here I suppose as well.

src/primitives/fieldvec.rs
```diff
@@ -0,0 +1,380 @@
...
+    /// field, via the inclusion map.
+    pub fn lift<E: Field + From<F>>(&self) -> FieldVec<E> {
+        self.iter().cloned().map(E::from).collect()
+    }
+}
+
+impl<F: Default> FieldVec<F> {
+    /// Pushes an item onto the end of the vector.
+    pub fn push(&mut self, item: F) {
+        self.len += 1;
```
tcharding, line 172: Perhaps its slightly cleaner to put this increment at the end of the function, saves all the `+ 1`s below and saves reader from having to check there are no return paths where the increment was done and an item not added.

src/primitives/fieldvec.rs
```diff
@@ -0,0 +1,380 @@
...
+        self.assert_has_data();
+
+        #[cfg(not(feature = "alloc"))]
+        {
+            self.inner_a[self.len - 1] = item;
+        }
```
tcharding, line 178:
```suggestion
        #[cfg(not(feature = "alloc"))]
        self.inner_a[self.len - 1] = item;
```

Or `self.inner_a[self.len] = item` if you use my other suggestion.

src/primitives/fieldvec.rs
tcharding: Oh, did you write this to be uniform with `pop`?

tcharding (commented):
There were a couple of other super trivial docs things I saw but we'll likely need to go over the docs with a fine tooth comb before we try to do 1.0 so I didn't bother mentioning them.

src/primitives/fieldvec.rs
clarkmoody, line 72: Should we have explicit tests for this type?

apoelstra: `pub(crate)` would make this visible to the entire crate. `pub` makes it visible only to the parent module (who in turn re-exports it privately to make it visible to other submodules).

I could do `pub(super)` if it makes you happier.

src/primitives/fieldvec.rs
apoelstra: Yeah, agreed, it'd be cleaner to move the increment. Will do.

src/primitives/fieldvec.rs
apoelstra: Sure, I'll add them. Though all these docs are private of course.

src/primitives/fieldvec.rs
apoelstra: Did you try compiling this suggestion?

src/primitives/fieldvec.rs
apoelstra: Yeah, I think it'd be worthwhile. In later commits I add some tests for things I was uncertain of but I should move them back and expand them.

apoelstra: FWIW every function in here is exercised in some form by the error correction logic, and I have run 15 billion fuzz iterations on it now on codex32, on codex32 without alloc (which always just fails immediately) and on bech32 without alloc. So I am confident in their correctness.

src/primitives/fieldvec.rs
apoelstra: @tcharding actually, the increment is before the `assert_has_data` assertion so that the panic message will be informative if you try to push too much data. Otherwise there'll just be an "index out of bounds" panic in the no-alloc case.

apoelstra:
Improved doccomments and added a bunch of unit tests. Also added missing `Index<Range<usize>>` impl on fieldvec.

apoelstra (commented):
Successfully ran local tests on c209ce8b12954102a7b0219a44ac09f7f0f7587e.

src/primitives/fieldvec.rs
tcharding: Interesting, I missed that. Thanks.

src/primitives/fieldvec.rs
tcharding: Oh with private `field` module, I've never thought to do that before. Thanks for the explanation.

src/primitives/fieldvec.rs
tcharding: Should have had two coffees, didn't notice the whole `field` module being private yesterday.

src/primitives/fieldvec.rs
tcharding: > error[E0658]: attributes on expressions are experimental

Worst review suggestion ever :)

tcharding (approved):
ACK c209ce8b12954102a7b0219a44ac09f7f0f7587e

src/primitives/fieldvec.rs
clarkmoody: I assumed that these were upstream of other functions being exercised, but it seemed odd not to have any unit tests on a Vec-like type.

clarkmoody (approved):
ACK c209ce8

Tests running. No further code comments.

PR #208, opened 2025-01-06
Correct internal `decode` docs

tnull:
Previously, the field docs stated that the `Hrp` would be always lowercase. However, we only assert that it's not mixed-case, i.e., all upper-case characters are allowed.

### DIFF
--- a/src/primitives/decode.rs
+++ b/src/primitives/decode.rs
@@ -114,7 +114,7 @@ const SEP: char = '1';
 /// ```
 #[derive(Debug)]
 pub struct UncheckedHrpstring<'s> {
-    /// The human-readable part, guaranteed to be lowercase ASCII characters.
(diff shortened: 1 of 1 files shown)


tcharding (approved):
ACK f7dd6e5ddf4cb0b14f21587137d0b0799c5ad404

tcharding:
I couldn't remember the upper/lower case logic so I checked the claim in the PR description and it looks correct to me.

Thanks @tnull

apoelstra (approved):
ACK f7dd6e5ddf4cb0b14f21587137d0b0799c5ad404; successfully ran local tests

clarkmoody (approved):
ACK f7dd6e5ddf4cb0b14f21587137d0b0799c5ad404

PR #211, opened 2025-01-29
Update LICENSE-MIT

maximevtush:
Updated the copyright year from 2017 to 2025 in the LICENSE file

### DIFF
--- a/LICENSE-MIT
+++ b/LICENSE-MIT
@@ -1,4 +1,4 @@
-Copyright (c) 2017 Clark Moody
+Copyright (c) 2025 Clark Moody
 
 Permission is hereby granted, free of charge, to any
 person obtaining a copy of this software and associated
--- a/LICENSE-MIT
+++ b/LICENSE-MIT
@@ -1,4 +1,4 @@
-Copyright (c) 2025 Clark Moody
+Copyright (c) 2017-2025 Clark Moody
 
 Permission is hereby granted, free of charge, to any
 person obtaining a copy of this software and associated


apoelstra (approved):
ACK 7ccf7e82b7d6cf609a3e53f49e18b68429b3a84d; successfully ran local tests; fine by me

tcharding:
Written ok from me in relation to any work I have done on this crate - if such a thing is needed.

clarkmoody:
According to [this random Stack Exchange answer](https://opensource.stackexchange.com/questions/1522/what-should-be-written-in-mit-license-year-full-name), the copyright year should be either the year the license went into effect or a range (e.g. `2017-2025`).

tcharding:
What is the benefit of this now? Do we have to update the 2025 every year? That seems like a step backwards.

tcharding:
Perhaps make a case for why this change is needed in the first place otherwise it looks to me like its not. I am not a lawer and have zero interest in copyright, I care about maintainability though.

tcharding:
I can't spell either.

apoelstra:
> Do we have to update the 2025 every year?

I think we should try to remember to do it every decade or so, as long as the project is maintained. In theory if we have a super old date such that a copyright from that date would've lapsed, somebody could claim they thought the work had fallen into the public domain.

But IANAL and I also really don't think this matters.

I would not advise deleting the year unless there is a clear reason for it.

tcharding:
Does the 2017 go stale some time? Like 10 years (2027)?

apoelstra:
I think it "goes stale" 70 years after Clark dies :P.

tcharding:
Mad, I'll be dead then too. Lets just leave it as it is then huh.

clarkmoody:
> I think it "goes stale" 70 years after Clark dies :P.

👀

clarkmoody:
Really though, I don't think this matters, and we should probably just close this PR.

PR #212, opened 2025-02-07
hrp len const fn

tzemanovic:
makes the `Hrp::len` a const fn

### DIFF
--- a/src/primitives/hrp.rs
+++ b/src/primitives/hrp.rs
@@ -263,7 +263,7 @@ impl Hrp {
     /// Guaranteed to be between 1 and 83 inclusive.
     #[inline]
     #[allow(clippy::len_without_is_empty)] // HRP is never empty.
-    pub fn len(&self) -> usize { self.size }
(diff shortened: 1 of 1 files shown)


clarkmoody (approved):
ACK 72ad45ac3a8ad5da31e545e431790c2d27d9987a

tcharding:
What is the motivation? Just because we can and its better? If so perhaps we do `FieldVec::len` at the same time. I checked all the other `len` functions and they are all `ExactSizedIterator` trait methods so cannot be const.

apoelstra:
Every pure function should be `const`, unless there are compiler limitations (such as them being trait methods) that prevent it, or we expect to run into compiler limitations with future refactorings.

apoelstra (approved):
ACK 72ad45ac3a8ad5da31e545e431790c2d27d9987a; successfully ran local tests

nazar-pc:
Could someone publish a new release with this change, please?

PR #219, opened 2025-07-07
Run the linter

tcharding:
Clippy emits a bunch of warnings of form:

 lifetime flowing from input to output with different syntax can be confusing

Fix them as suggested by using the anonymous lifetime.


tcharding:
> Thanks! Can you also run the formatter?

Can do.

> If you're really motivated I think it'd be ok to migrate the whole rust-bitcoin CI infrastructure here. (I've done it in https://github.com/ElementsProject/rust-elements/ say if you want a simpler case to copy)

Hot potato - cc @jamillambert?

PR #221, opened 2025-07-25
Implement From<ChecksumError> for DecodeError

stevenroose:
Noticed this was missing.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -495,6 +495,12 @@ impl From<UncheckedHrpstringError> for DecodeError {
     fn from(e: UncheckedHrpstringError) -> Self { Self::Parse(e) }
 }
 
+#[cfg(feature = "alloc")]
+impl From<ChecksumError> for DecodeError {
+    #[inline]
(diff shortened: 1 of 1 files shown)


apoelstra:
Yeah, concept ACK being consistent here. We should have one or both. Will assume that we should go with "both" unless @clarkmoody steps in and says otherwise.

But personally I've been moving away from these `From` impls. It's not a lot of extra code to add `.map_err(DecodeError::Checksum)` before `?`, makes it waaay easier to grep for (and `git log -S` for) where each variant is used, and it keeps the public API tighter. 

I also increasingly think that `?` being a single character was a mistake for Rust, because it makes it incredibly easy to propagate errors without adding any useful context. If I'm already using `.map_err` everywhere it's a smaller difference between adding context and not.

clarkmoody:
Concept ACK from me.

Personally, I like the `?` operator 😉

apoelstra:
concept ACK -- if you rebase on #223 I can do a tested ACK.

PR #220, opened 2025-07-25
Remove unused argument on CheckedHrpstring::fe32_iter

stevenroose:
Trying to solve https://github.com/rust-bitcoin/rust-bech32/pull/193#issuecomment-3117576664

PR #223, opened 2025-08-01
Fix CI and pin nightly so it shouldn't break again

apoelstra:
Right now I can't test #220 or #221 locally because clippy is failing with the latest nightly.

I can disable the local clippy check but I'd prefer to fix it here.

Should unblock both those PRs.


apoelstra:
Something is broken with GA. Will try again in a few hours or days.

clarkmoody (approved):
ACK 7a3f493e58fe28035b8a2d56a092a7359328e0f0

Tests passing on Rust 1.88

PR #234, opened 2026-01-16
Re-write CI to use `cargo-rbmt`

tcharding:
Re-write CI to use the `cargo-rbmt` tool [0]

This is quite an overhaul, some notes:

- Adds action to install `cargo-rbmt` directly from the repo
- Upgrades to checkout v5 (and uses explicit commit hash)
- Uses ubuntu-24.04 instead of latest (I can't remember why we did this)
- Defaults jobs to no permissions and uses read in each job
- Gets the MSRV from manifest instead of explicitly
- Removes the `Beta` toolchain test

[0] https://github.com/rust-bitcoin/rust-bitcoin-maintainer-tools/tree/master/cargo-rbmt


apoelstra:
FYI I'm happy to review this if we can get it working.

This crate is currently pretty annoying for me to test because it doesn't have embedded lockfiles.

tcharding:
Replaced by #254

PR #253, opened 2026-04-15
primitives: fix min/max size hint in HrpFe32Iter

satsfy:
While running mutation tests, I noticed a defect.

`HrpFe32Iter` models the BIP 173 HRP expansion used for checksum input.

In [BIP 173](https://en.bitcoin.it/wiki/BIP_0173), HRP expansion is: `[ord(x) >> 5 for x in hrp] + [0] + [ord(x) & 31 for x in hrp]`

In `size_hint()`, the high branch already includes that single inserted separator element. Adding another +1 in the final sum overcounts the remaining length by one.

This patch removes the extra +1 from the final min and max calculations and adds a test asserting that `size_hint()` matches the exact remaining length as the iterator is drained.

This change only corrects iterator length accounting. It does not change the yielded sequence of field elements or checksum semantics.

### DIFF
--- a/src/primitives/checksum.rs
+++ b/src/primitives/checksum.rs
@@ -502,8 +502,8 @@ impl Iterator for HrpFe32Iter<'_> {
             None => (0, Some(0)),
         };
 
-        let min = high.0 + 1 + low.0;
-        let max = high.1.zip(low.1).map(|(high, low)| high + 1 + low);
+        let min = high.0 + low.0;
+        let max = high.1.zip(low.1).map(|(high, low)| high + low);
 
         (min, max)
     }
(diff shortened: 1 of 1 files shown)


apoelstra:
Nice! Win for mutation testing. Thanks for writing a test that covers this.

apoelstra (approved):
ACK ca024f940d4ad47f40563eda7b6bf75f735fed5c; successfully ran local tests

clarkmoody (approved):
utACK ca024f940d4ad47f40563eda7b6bf75f735fed5c

tcharding (approved):
ACK ca024f940d4ad47f40563eda7b6bf75f735fed5c

PR #254, opened 2026-04-15
ci: migrate to cargo rbmt

satsfy:
This PR migrates CI from `contrib/test.sh` to `cargo rbmt`. It continues PR https://github.com/rust-bitcoin/rust-bech32/pull/234.

It also:
- adds minimal and recent lockfiles for CI
- declares `rust-version = "1.56.1"` in `Cargo.toml`
- adds rbmt toolchain metadata
- removes the old `contrib/test.sh` workflow
- updates the workflow to use the pinned upstream `setup-rbmt` action
- simplifies the embedded jobs into a matrix
- drops the stale mutagen lint override

No library behavior is intended to change.


apoelstra:
In 6cab5af2f89160e417a069c8c8a7b34e26f2a482:

The commit description says this "Gets the MSRV from manifest instead of explicitly" but actually it removes the MSRV from Cargo.toml, ignores the MSRV in fuzz/Cargo.toml, adds it explicitly in a Github Action file which my local CI system can't read, and doesn't add it anywhere else.

It also introduces lockfiles with version 4 which I'm fairly sure Cargo 1.56 can't read, so I don't think we're actually testing the MSRV here.

apoelstra:
This is the first commit which is authored by @tcharding.

tcharding:
Excluding the fact that I just made mistakes `cargo rbmt` has changed a lot since I did #234. I don't think you should use my patch as it is, instead just roll it into your own work. I would have thought this PR would be a single patch.

satsfy:
Squashed to a single commit, as suggested.

tcharding:
Can you grab the latest version of `cargo rbmt` mate? There shouldn't be a `rbmt.toml` file anymore. Just take a look at `rust-bitcoin` to see examples.

satsfy:
Sorry it really was not the most up-to-date version of rbmt. 

Force push removed `rbmt.toml` and use latest rbmt commit [af3c28](https://github.com/rust-bitcoin/rust-bitcoin-maintainer-tools/commits/master/cargo-rbmt).

Additionally, do you think I should read the version from `rbmt-version` file for CI, so that it does not repeat itself?

tcharding:
> Additionally, do you think I should read the version from rbmt-version file for CI, so that it does not repeat itself?

hmm, @nyonson I'm a bit confused about `rbmt-version`. In `rust-bitcoin` is the plan to remove it completely or keep using it in the cron jobs and justfile/contrib scripts? Or said another way, can you advise on the best move here?

apoelstra:
(Reminder to self that when this resolves, to copy its approach into rust-elements.)

nyonson:
> > Additionally, do you think I should read the version from rbmt-version file for CI, so that it does not repeat itself?
> 
> hmm, @nyonson I'm a bit confused about `rbmt-version`. In `rust-bitcoin` is the plan to remove it completely or keep using it in the cron jobs and justfile/contrib scripts? Or said another way, can you advise on the best move here?

I assume @satsfy you are talkin about the uses lines in the yaml? Like `- uses: rust-bitcoin/rust-bitcoin-maintainer-tools/.github/actions/setup-rbmt@af3c2868415b17eedc808da6d0589e10b7482660 # v0.1.0`? If you find a clean way to read the rbmt version into that, I'd love to see it. I have struggled with Github Actions because I don't think the `uses` key allows much to be run in it. So I have just been using this totally annoying strategy of copying the hash everywhere.

@tcharding I have been trying to think of a good way to merge these version definitions, but haven't found a good pattern yet. Top of the list for rbmt stuff though.

.github/workflows/rust.yml
```diff
@@ -2,133 +2,139 @@ on: [push, pull_request]
...
-        run: echo "nightly_version=$(cat nightly-version)" >> $GITHUB_OUTPUT
+      - uses: rust-bitcoin/rust-bitcoin-maintainer-tools/.github/actions/setup-rbmt@af3c2868415b17eedc808da6d0589e10b7482660 # v0.1.0
+      - uses: Swatinem/rust-cache@v2
+      - name: Run ${{ matrix.toolchain }} tests
+        run: cargo rbmt --lock-file ${{ matrix.dep }} test --toolchain ${{ matrix.toolchain }}
 
-  Stable:
-    name: Test - stable toolchain
-    runs-on: ubuntu-latest
+  MSRV: # 2 jobs: 1 toolchain × 2 lock files.
```
nyonson: I think you can drop this `MSRV` job and just extend the `Test` matrix above with `toolchain: [stable, nightly, msrv]`. Kinda related, but we have also talked about in other repos how testing on nightly isn't all that helpful.

satsfy: Thanks, its done now. MSRV is now part of the test matrix and the separate job is gone. Need to use a MSRV-specific command path because `cargo rbmt --toolchain msrv` currently breaks on Rust 1.56/1.57 metadata parsing.

About nightly, the CI already had a nightly test so I'm trying to retain it as is for this PR.

satsfy:
About the yaml file version redeclarations, the best alternative solution I found would be adding a new github actions file just for reading it from file. I don't think the added complexity of this solution would be worth it, so I'm leaving that unchanged.

One possibility here is to point to master instead of point to a commit. I suppose this should be possible: `rust-bitcoin/rust-bitcoin-maintainer-tools/.github/actions/setup-rbmt@master # v0.1.0` (but this wouldn't be stable) or `rust-bitcoin/rust-bitcoin-maintainer-tools/.github/actions/setup-rbmt@rbmtv0.1.0` if `rust-bitcoin-maintainer-tools` had this custom branch.

satsfy:
Latest commit fixes lint errors by constraining feature flags.

tcharding (approved):
ACK 30686ad7e22cb30ff5b6a95cd891756b777a96ce

tcharding:
Looks good to me! PR description is stale mate, still mentions `rbmt.toml`.

apoelstra (approved):
ACK 30686ad7e22cb30ff5b6a95cd891756b777a96ce; successfully ran local tests

.github/workflows/rust.yml
nyonson: I'll try to hammer out the MSRV issue over here: https://git.rust-bitcoin.org/rust-bitcoin/rust-bitcoin-maintainer-tools/issues/126

apoelstra:
Merged this but FYI it does not update the weekly cronjob, so we are updating the nightly version in the wrong place. (Noticed when porting this to rust-elemnets.)

PR #255, opened 2026-04-17
test: address all mutation test misses and timeouts

### LINKED ISSUE #252: New Mutants Found
Displaying up to the first 10 mutants:
```
src/lib.rs:364:5: replace encode_to_writer -> Result<(), EncodeIoError> with Ok(())
src/lib.rs:470:9: replace <impl fmt::Display for DecodeError>::fmt -> fmt::Result with Ok(Default::default())
src/lib.rs:482:9: replace <impl std::error::Error for DecodeError>::source -> Option<&(dyn std::error::Error +'static)> with None
src/lib.rs:515:9: replace <impl fmt::Display for EncodeError>::fmt -> fmt::Result with Ok(Default::default())
src/lib.rs:527:9: replace <impl std::error::Error for EncodeError>::source -> Option<&(dyn std::error::Error +'static)> with None
src/lib.rs:560:9: replace <impl fmt::Display for EncodeIoError>::fmt -> fmt::Result with Ok(Default::default())
src/lib.rs:572:9: replace <impl std::error::Error for EncodeIoError>::source -> Option<&(dyn std::error::Error +'static)> with None
src/segwit.rs:135:5: replace encode_v1 -> Result<String, EncodeError> with Ok(String::new())
src/segwit.rs:135:5: replace encode_v1 -> Result<String, EncodeError> with Ok("xyzzy".into())
src/segwit.rs:218:21: replace += with -= in encode_upper_to_fmt_unchecked
```
Running cargo mutants version: cargo-mutants 27.0.0
For the complete list, please check the [mutants.out artifact](https://github.com/rust-bitcoin/rust-bech32/actions/runs/24296848754).

satsfy:
Closes https://github.com/rust-bitcoin/rust-bech32/issues/252
Closes https://github.com/rust-bitcoin/rust-bech32/issues/250

Out of 1254 mutation tests the CI runs weekly, 316 were misses and 28 timed out. This PR addresses all misses and timeouts by adding 39 new tests and refining 1 existing test. 

This PR also updates `.cargo/mutants.toml` to exclude 74 cases from the weekly cargo mutants run. Each exclusion is documented inline and falls into one of four buckets: equivalent mutations in characteristic-2 arithmetic, code that is not exercised in this configuration, mutations that only affect representation or allocation without changing observable behavior, and mutations that only turn loops or iterators into non-terminating cases.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -748,8 +748,9 @@ mod tests {
 
         match encode::<Bech32m>(hrp, &data) {
             Ok(_) => panic!("false positive"),
-            Err(EncodeError::TooLong(CodeLengthError { encoded_length, code_length: _ })) =>
-                assert_eq!(encoded_length, 1024),
+            Err(EncodeError::TooLong(CodeLengthError { encoded_length, code_length: _ })) => {
+                assert_eq!(encoded_length, 1024)
+            }
             _ => panic!("false negative"),
         }
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -593,6 +593,9 @@ impl From<std::io::Error> for EncodeIoError {
 #[cfg(test)]
 #[cfg(feature = "alloc")]
 mod tests {
+    #[cfg(feature = "std")]
+    use std::error::Error;
+
     use super::*;
 
     // Tests below using this data, are based on the test vector (from BIP-173):
@@ -603,6 +606,45 @@ mod tests {
         0x54, 0x94, 0x1c, 0x45, 0xd1, 0xb3, 0xa3, 0x23,
         0xf1, 0x43, 0x3b, 0xd6,
     ];
+
+    #[test]
+    fn errors_have_non_empty_display_and_source() {
+        let decode_err = decode("test1lu08d6qejxtdg4y5r3zarvary0c5xw7kw79nny")
+            .expect_err("checksum error expected");
+        assert!(!decode_err.to_string().is_empty());
+
+        let too_long_data = [0_u8; 632];
+        let encode_err = encode::<Bech32m>(Hrp::parse_unchecked("abcde"), &too_long_data)
+            .expect_err("too long error expected");
+        assert!(!encode_err.to_string().is_empty());
+
+        #[cfg(feature = "std")]
+        {
+            assert!(decode_err.source().is_some());
+            assert!(encode_err.source().is_some());
+
+            struct BrokenWriter;
+            impl std::io::Write for BrokenWriter {
+                fn write(&mut self, _buf: &[u8]) -> std::io::Result<usize> {
+                    Err(std::io::Error::new(
+                        std::io::ErrorKind::Other,
+                        "mutation regression test write failure",
+                    ))
+                }
+
+                fn flush(&mut self) -> std::io::Result<()> { Ok(()) }
+            }
+
+            let io_err = encode_to_writer::<Bech32, _>(
+                &mut BrokenWriter,
+                Hrp::parse_unchecked("test"),
+                &DATA,
+            )
+            .expect_err("writer error expected");
+            assert!(!io_err.to_string().is_empty());
+            assert!(io_err.source().is_some());
+        }
+    }
 
     #[test]
     fn encode_bech32m() {
--- a/src/primitives/checksum.rs
+++ b/src/primitives/checksum.rs
@@ -4,8 +4,6 @@
 //!
 //! [BCH]: <https://en.wikipedia.org/wiki/BCH_code>
 
-#[cfg(all(feature = "alloc", not(feature = "std")))]
-use alloc::vec::Vec;
 use core::marker::PhantomData;
 use core::{fmt, mem, ops};
 
@@ -512,6 +510,8 @@ impl Iterator for HrpFe32Iter<'_> {
 #[cfg(test)]
 mod tests {
     #[cfg(feature = "alloc")]
+    use alloc::vec::Vec;
+    #[cfg(feature = "alloc")]
     use core::convert::TryFrom;
 
     use super::*;
@@ -598,6 +598,50 @@ mod tests {
         .to_string();
         #[cfg(feature = "std")]
         println!("{}", _s);
+    }
+
+    #[test]
+    #[should_panic]
+    fn sanity_check_panics_on_bad_checksums() {
+        enum BadGeneratorShifts {}
+        impl Checksum for BadGeneratorShifts {
+            type MidstateRepr = u32;
+            type CorrectionField = crate::Fe1024;
+            const ROOT_GENERATOR: Self::CorrectionField = crate::Fe1024::new([Fe32::P, Fe32::X]);
+            const ROOT_EXPONENTS: core::ops::RangeInclusive<usize> = 24..=26;
+            const CODE_LENGTH: usize = 1023;
+            const CHECKSUM_LENGTH: usize = 6;
+            const GENERATOR_SH: [u32; 5] = [1, 1, 1, 1, 1];
+            const TARGET_RESIDUE: u32 = 1;
+        }
+
+        BadGeneratorShifts::sanity_check();
+    }
+
+    #[test]
+    #[cfg(feature = "alloc")]
+    fn print_impl_renders_expected_bech32_impl() {
+        let rendered = PrintImpl::<crate::Fe1024>::new(
+            "Bech32",
+            &[Fe32::A, Fe32::K, Fe32::_5, Fe32::_4, Fe32::A, Fe32::J],
+            &[Fe32::Q, Fe32::Q, Fe32::Q, Fe32::Q, Fe32::Q, Fe32::P],
+        )
+        .to_string();
+
+        let generator_shift = "0x4eb8406b0"; // generator << 2^3
+        assert!(
+            rendered.contains(generator_shift),
+            "generated impl missing expected fragment: {}\n\n{}",
+            generator_shift,
+            rendered
+        );
+    }
+
+    #[test]
+    fn packed_null() {
+        let mut packed = PackedNull;
+        assert_eq!(packed.unpack(0), 0);
+        assert_eq!(packed.mul_by_x_then_add(1, 31), 0);
     }
 
     #[test]
--- a/src/primitives/gf32.rs
+++ b/src/primitives/gf32.rs
@@ -390,6 +390,9 @@ impl From<Infallible> for TryFromError {
 
 #[cfg(test)]
 mod tests {
+    #[cfg(feature = "std")]
+    use core::convert::TryFrom;
+
     use super::*;
 
     #[test]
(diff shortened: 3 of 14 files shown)


src/lib.rs
```diff
@@ -604,6 +604,39 @@ mod tests {
         0xf1, 0x43, 0x3b, 0xd6,
     ];
 
+    #[test]
+    #[cfg(feature = "std")]
+    fn errors_have_non_empty_display_and_source() {
+        let decode_err = decode("test1lu08d6qejxtdg4y5r3zarvary0c5xw7kw79nny")
+            .expect_err("checksum error expected");
+        assert!(!decode_err.to_string().is_empty());
+        assert!(std::error::Error::source(&decode_err).is_some());
```
tcharding: Can we just have the feature gate on the `source` function call? And I rekon cleaner to import `std::error`.

src/lib.rs
```diff
@@ -706,8 +739,9 @@ mod tests {
 
         match encode::<Bech32m>(hrp, &data) {
             Ok(_) => panic!("false positive"),
-            Err(EncodeError::TooLong(CodeLengthError { encoded_length, code_length: _ })) =>
-                assert_eq!(encoded_length, 1024),
+            Err(EncodeError::TooLong(CodeLengthError { encoded_length, code_length: _ })) => {
+                assert_eq!(encoded_length, 1024)
```
tcharding, line 752: Looks like a formatting change?

src/primitives/checksum.rs
```diff
@@ -513,6 +513,7 @@ impl Iterator for HrpFe32Iter<'_> {
 mod tests {
     #[cfg(feature = "alloc")]
     use core::convert::TryFrom;
+    use std::panic;
```
tcharding: I'm surprised this builds without a feature gate on `std`.

src/primitives/checksum.rs
```diff
@@ -599,6 +600,65 @@ mod tests {
...
+            const CODE_LENGTH: usize = 1023;
+            const CHECKSUM_LENGTH: usize = 6;
+            const GENERATOR_SH: [u32; 5] = [1, 1, 1, 1, 1];
+            const TARGET_RESIDUE: u32 = 1;
+        }
+
+        assert!(
+            panic::catch_unwind(BadGeneratorShifts::sanity_check).is_err(),
+            "sanity_check should panic on inconsistent generator shifts"
+        );
```
tcharding: Where did you get this pattern from mate? There is `should_panic` attribute that is already used in this repo.

src/primitives/gf32.rs
```diff
@@ -390,6 +390,8 @@ impl From<Infallible> for TryFromError {
 
 #[cfg(test)]
 mod tests {
+    use core::convert::TryFrom;
```
tcharding, line 394: I think thing will trigger a lint warning for unused_import when building with `std`. Could use if not `alloc` enabled feature gate. Just a thought, did not test.

src/primitives/gf32.rs
```diff
@@ -510,6 +512,54 @@ mod tests {
...
+
+    #[test]
+    #[cfg(feature = "std")]
+    fn conversion_error_display_and_source() {
+        let invalid = Fe32::from_char('b').unwrap_err();
+        assert!(invalid.to_string().contains("invalid"));
+
+        let not_a_byte = Fe32::try_from(256u32).unwrap_err();
+        assert!(not_a_byte.to_string().contains("invalid field element"));
+        assert!(std::error::Error::source(&not_a_byte).is_some());
```
tcharding, line 552: We don't guarantee stability of error `Display` impls in general so instead we typically (in `rust-bitcoin` anyway) test like this

```rust
                assert!(!err.to_string().is_empty());
                #[cfg(feature = "std")]
                assert!(err.source().is_some());
```

src/primitives/hrp.rs
```diff
@@ -490,6 +490,8 @@ impl std::error::Error for Error {
 
 #[cfg(test)]
 mod tests {
+    use core::hash::{Hash, Hasher};
```
tcharding, line 493: Is this needed?

tcharding:
Top effort man! I didn't get to review the excludes, gotta run.

Reviewed upto: a041f3642a78fd14ec4992bbe507cd755863ac90

luisschwab:
@satsfy FIY: If you do "Closes #<issue_number>" GitHub will auto-close the issue upon merging.

tcharding:
I have no idea if the excludes are correct because I can't read all the comments because they are too math heavy for me. Did you write them yourself or did you use an LLM? Have you a math major? 

If you have a math major and/or wrote them yourself I rekon @apoelstra might review them. If you used an LLM to write them then its probably too demanding of review time.

tcharding (approved):
ACK <elided> - ok to merge but I have no clue if the exclude comments are correct.

EDIT: Rescind ack due to comments below.

satsfy:
I used LLM to generate the excludes initially, cut down a part, then challenged a sample for each group of excludes, but each one appeared sensible.

> In a field of characteristic two, everything is its own additive inverse, so x=−x for every element

Source: https://math.stackexchange.com/a/1573310

This peculiarity for characteristic 2 fields explains a good portion of the excludes.

The timeout paths are very straightforward in comparison, a quick code read explains them. I could extend a fix like [a041f36](https://github.com/rust-bitcoin/rust-bech32/pull/255/commits/a041f3642a78fd14ec4992bbe507cd755863ac90) to tackle more of them, but this comes a the cost of modifying the lib instead of building a test around it.

Do note that non-`all-features` builds are excluded from mutation tests, because it only runs with `all-fields` flag. Do you think I should modify this PR to handle testing all codebase even the excluded paths too? If not, I guess this would be a good follow up (that i forgot to mention).

Suppose we could keep going and attempt to remove some more excludes, but the PR size balloons fast. Let me know what you think.

src/lib.rs
satsfy: Isolated into its own formatting commit

src/primitives/hrp.rs
satsfy: About `Hash`, when I remove it says: 
>"no method named `hash` found for struct `primitives::hrp::Hrp` in the current scope items from traits can only be used if the trait is in scope"

Regarding: 
```rs
x.hash(&mut h1);
```
Meanwhile `Hasher` is directly used below.

tcharding:
Thanks man. I happy with your work and happy to merge. Leaving for @apoelstra to comment on any math stuff if needed. Cheers mate.

apoelstra:
In 5b4a9dd6d9284e6a8cab36805bc2cdc40ab76970:

Can you just put a comment in there saying that these are LLM-generated? These all look right to me but I don't think it'd be a good use of effort for me to carefully reason through every one of them.

If you were **really** motivated you could add a non-char-2 field to the crate and see if things still work :P. But the answer is probably no, actually, and that's fine. Char 2 lets us take a lot of shortcuts and some of them are algorithmically meaningful, not just "I don't care to keep track of parity" or "I'd like to hit whichever of the `+` and `-` keys are closer to my fingers right now".

apoelstra:
@tcharding on a high level there are a few sources of mutations here:

* in characteristic-2 fields, addition and subtraction and xor are all the same operation, so you can swap these operations and nothing will change
* we're lazy about detecting and producing trailing zeros in our polynomials
* we have some weird no-alloc stuff in here where we have codepaths that can't actually be hit if you're compiling with alloc on
* some stuff about ascii and our character set
* the usual `Vec`-capacity-type things that don't affect code correctness

The one that raises my eyebrows is:

* the `Current BCH/LFSR fixtures do not distinguish multiply vs divide in these update terms`

I think this one might actually be wrong (we should be able to test this). Maybe we should move this to its own section with a note saying that we should try to revisit it.

tcharding:
Thanks man! Removed my ack to give @satsfy time.

src/primitives/checksum.rs
satsfy: Using `should_panic` now

src/primitives/gf32.rs
satsfy: Gated it with `#[cfg(feature = "std")]`

src/primitives/gf32.rs
satsfy: Every such instance has been modified to conform.

src/primitives/checksum.rs
satsfy: Using `should_panic` now so this import was dropped.

src/lib.rs
satsfy: Now only the lines that truly require `std` are conditionally compiled, with `std::error::Error` imported only in that `std` case.

satsfy:
In the latest force push I addressed all comments, included on additional commit formatting change and fixed flags.  

Was able to remove the exclude that mutation `Current BCH/LFSR fixtures do not distinguish multiply vs divide in these update terms` by using LLM to generate various hypothesis, testing ~1200 mutations and keeping the asserts that matter. A branch was not being tested in the Berlekamp-Massey algorithm.

Mutation `src/primitives/lfsr.rs:121:55: replace * with / in LfsrIter<F>::berlekamp_massey` is handled by:
```rs
let step5_sequence: Vec<_> =
    LfsrIter::berlekamp_massey(&[Fe32::A, Fe32::C, Fe32::A, Fe32::Y, Fe32::A])
        .take(4)
        .collect();
assert_eq!(step5_sequence, [Fe32::A, Fe32::C, Fe32::A, Fe32::Y]);
``` 

And mutation `src/primitives/lfsr.rs:131:55: replace * with / in LfsrIter<F>::berlekamp_massey` is handled by:
```rs
let sequence: Vec<_> =
    LfsrIter::berlekamp_massey(&[Fe32::Y, Fe32::H, Fe32::Q, Fe32::Q]).take(10).collect();
assert!(sequence[2..].iter().all(|&fe| fe == Fe32::Q));
```  

Learned more about the domain by watching some tutorials on fields and the algorithm, but I have to be honest - I'm shaky at best. The positive outcome here is that we don't have excludes for LFSR anymore. Do you think these are good enough?

src/primitives/hrp.rs
tcharding: oooph, hard to good help huh. Thanks mate.

tcharding (approved):
ACK 72dd5cc161c5caa895551a787ae30b2919f6ef84

src/primitives/polynomial.rs
```diff
@@ -102,7 +102,7 @@ impl<F: Field> Polynomial<F> {
     /// Helper function to add leading 0 terms until the polynomial has a specified
     /// length.
     fn zero_pad_up_to(&mut self, len: usize) {
-        while self.inner.len() < len {
+        for _ in self.inner.len()..len {
```
tcharding, line 105: In a587b8e343a840a71ae4dc5825278205e864d6c2 I didn't understand the commit log 'bound ...'. Also I don't know exactly what either of these code blocks compiles down to. I'm guessing the iterator version may be faster? FWIW the code is slightly harder to read but I _think_ this is an improvement.

tcharding (approved):
ACK cc6593f032c0690a2b8b19f57fda87dab659ebd6

src/primitives/polynomial.rs
apoelstra: I would think/hope they compile to the same thing. But the stupid mutation tester wants to screw with `<` but can't screw with `..` I guess, which is an interesting reason to do this..

apoelstra (approved):
ACK cc6593f032c0690a2b8b19f57fda87dab659ebd6; successfully ran local tests

PR #259, opened 2026-04-22
ci: unify nightly versions in Cargo.toml

satsfy:
Follow up from https://github.com/rust-bitcoin/rust-bech32/pull/254#issuecomment-4289786340

After the rbmt CI migration, most jobs were already consuming nightly from `[package.metadata.rbmt.toolchains]` in `Cargo.toml`, but the weekly updater still wrote `nightly-version`. This change makes `Cargo.toml` the single nightly source of truth across the updater and CI jobs.

Changes:
- update weekly nightly cron to use `cargo rbmt toolchains --update-nightly.` Replicates rust-bitcoin CI closely.
- `nightly-version` removed.
- migrate Embedded CI job to read nightly from `Cargo.toml` instead of `nightly-version.`

Validated in personal fork CI: 
- The CI run (didn't create the PR because it wasn't allowed): https://github.com/satsfy/rust-bech32/actions/runs/24799694669/job/72578713007
- PR produced: https://github.com/satsfy/rust-bech32/compare/master...satsfy:rust-bech32:create-pull-request/weekly-nightly-update?expand=1


.github/workflows/cron-weekly-update-nightly.yml
```diff
@@ -9,16 +9,13 @@ jobs:
...
+      - name: Update nightly toolchain pins
         run: |
           set -x
-          # Not every night has a nightly, so extract the date from whatever
-          # version of the compiler dtolnay/rust-toolchain gives us.
-          NIGHTLY_DATE=$(rustc +nightly --verbose --version | sed -ne 's/^commit-date: //p')
-          # Update the nightly version in the reference file.
-          echo "nightly-${NIGHTLY_DATE}" > nightly-version
-          echo "nightly_date=${NIGHTLY_DATE}" >> $GITHUB_ENV
+          eval "$(cargo rbmt toolchains --update-nightly)"
```
apoelstra: In 85595555b1a1efec3432421b73ac3fbe2b0394e2:

Why run `eval $(x)` instead of `x`?

apoelstra: I see, we're doing this in rust-bitcoin too...lol who is reviewiing this stuff?

apoelstra: Anyway this is fine. The code originates in https://github.com/rust-bitcoin/rust-bitcoin/pull/5869 fwiw

apoelstra: ...oh, but this doesn't copy the files from rust-bitcoin. It's new code.

apoelstra:
Can you just copy the files from rust-bitcoin so that I can diff them rather than re-reviewing? You might need to change the token but that should be it.

.github/workflows/cron-weekly-update-nightly.yml
satsfy: Seems like [rust-bitcoin also uses the eval](https://github.com/rust-bitcoin/rust-bitcoin/blob/fcaa17683bc7b42ea0457a331197046cf0b228fd/.github/workflows/cron-weekly-update-nightly.yml#L23), but I removed and [ran the CI without eval](https://github.com/satsfy/rust-bech32/actions/runs/24801512572/job/72585063229) and it [worked](https://github.com/satsfy/rust-bech32/compare/master...satsfy:rust-bech32:create-pull-request/weekly-nightly-update?expand=1).

I'll be combining this change with the rust-bitcoin nightly version to make it easier to review.

satsfy:
It now mirrors [rust-bitcoin's version](https://github.com/rust-bitcoin/rust-bitcoin/blob/fcaa17683bc7b42ea0457a331197046cf0b228fd/.github/workflows/cron-weekly-update-nightly.yml#L23) in everything but 1. absent `eval` and 2. the repo specific token. 

Validated again with [this CI run](https://github.com/satsfy/rust-bech32/actions/runs/24803053104/job/72590372894) creating [this branch](https://github.com/satsfy/rust-bech32/compare/master...satsfy:rust-bech32:create-pull-request/daily-nightly-update?expand=1) in my local CI.

Side question: Why don't we have a central CI repo for reusable workflows and shared actions (maybe just import from rust-bitcoin directly), and keep a tiny workflow file in each repo that only calls into that central logic?

.github/workflows/rust.yml
```diff
@@ -99,9 +99,9 @@ jobs:
         uses: actions/checkout@v4
       - name: Set up QEMU
         run: sudo apt-get update -y && sudo apt-get install -y qemu-system-arm gcc-arm-none-eabi
-      - name: Read nightly version
+      - name: Read nightly version from Cargo.toml
         id: nightly
-        run: echo "version=$(cat nightly-version)" >> $GITHUB_OUTPUT
+        run: echo "version=$(awk -F'\"' '/^nightly = / {print $2; exit}' Cargo.toml)" >> "$GITHUB_OUTPUT"
```
satsfy: Seems like [rust-bitcoin doesn't have this problem](https://github.com/rust-bitcoin/rust-bitcoin/blob/fcaa17683bc7b42ea0457a331197046cf0b228fd/.github/workflows/rust.yml#L131).

Edit: well it uses a clearer way of reading... I'll make this clearer by using `cargo metadata --format-version 1 --no-deps | jq -r '.packages[0].metadata.rbmt.toolchains.nightly'`

nyonson:
> Side question: Why don't we have a central CI repo for reusable workflows and shared actions (maybe just import from rust-bitcoin directly), and keep a tiny workflow file in each repo that only calls into that central logic?

I have been attempting that with the actions in rust-bitcoin-maintainer-tools, but been finding them to be more trouble than they are worth since rbmt contains the bulk of the logic. Might be easier for repos to just copy/pasta [something like this](https://git.rust-bitcoin.org/rust-bitcoin/rust-bitcoin-maintainer-tools/src/branch/master/actions/setup-rbmt/action.yml) into their own repo since it won't change much.

apoelstra (approved):
ACK b8a32c988e389ee7941838e98644041224f6f125; successfully ran local tests

tcharding:
Since you already diff'ed I can't add anything by reviewing this. But FWIW concept ACK.

PR #261, opened 2026-05-09
Automated daily update to rustc (to )

apoelstra:
Automated update to Cargo.toml workspace metadata by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ edition = "2018"
 rust-version = "1.56.1"
 
 [package.metadata.rbmt.toolchains]
-nightly = "nightly-2026-05-01"
+nightly = "nightly-2026-05-08"
 stable = "stable"
 
 [features]


tcharding (approved):
ACK 8001cd6ba8e665a970b5ac7095be1b350559f232

apoelstra:
derp.

The reason we are not getting the right nightly-version here is that in our cronjob we replaced

```
eval "$(cargo rbmt toolchains --update-nightly)"
```
with
```
cargo rbmt toolchains --update-nightly
```

because I complained about the use of `eval` here https://github.com/rust-bitcoin/rust-bech32/pull/259#discussion_r3126690276

Well, I'm an idiot. `x` just runs `x` while `eval "$(x)"` runs `x` and then *also runs the output of `x`*, which in this case sets some envvars which are later used.

I should've asked a chatbot instead of just assuming the original code was a dumb clankerism.

apoelstra:
ACK 8001cd6ba8e665a970b5ac7095be1b350559f232; successfully ran local tests

PR #262, opened 2026-05-12
ci: fix nightly cronjob to set environment variables properly

apoelstra:
We were not evaluating the output of `cargo-rbmt` because I was being dumb when we introduced the cronjob. As a result, our weekly rustc update PRs have bad titles and commit descriptions.


apoelstra:
On 07aba28aba53e0e3ace6e719f9e68b2cde0d7277 successfully ran local tests

tcharding (approved):
ACK 07aba28aba53e0e3ace6e719f9e68b2cde0d7277

satsfy (approved):
ACK 07aba28aba53e0e3ace6e719f9e68b2cde0d7277

In line with previous solution.

PR #265, opened 2026-05-16
Automated daily update to rustc (to nightly-2026-05-15)

apoelstra:
Automated update to Cargo.toml workspace metadata by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action

PR #266, opened 2026-05-21
Update to rustc nightly-2026-05-15 with lint fix

satsfy:
Supersedes https://github.com/rust-bitcoin/rust-bech32/pull/265

Daily update to Cargo.toml workspace metadata removing an empty return statement, which was being flagged by the linter.

### DIFF
--- a/src/primitives/decode.rs
+++ b/src/primitives/decode.rs
@@ -495,7 +495,7 @@ impl<'s> CheckedHrpstring<'s> {
         let padding_len = fe_iter.len() * 5 % 8;
 
         if padding_len > 4 {
-            return Err(PaddingError::TooMuch)?;
+            return Err(PaddingError::TooMuch);
         }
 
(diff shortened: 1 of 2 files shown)


src/primitives/decode.rs
```diff
@@ -495,7 +495,7 @@ impl<'s> CheckedHrpstring<'s> {
         let padding_len = fe_iter.len() * 5 % 8;
 
         if padding_len > 4 {
-            return Err(PaddingError::TooMuch)?;
+            Err(PaddingError::TooMuch)?;
         }
```
tcharding, line 499: This change is wrong. I'm intentionally not saying why ...

satsfy: That was dumb of me, I left it harder to read. Restored `return` and dropped `?` operator.

apoelstra (approved):
ACK 3a91a833c17f662e3c51647df37f1f9b98e2fcb6; successfully ran local tests

PR #277, opened 2026-06-16
fix: mixed-case validation in hrp::parse_display

### LINKED ISSUE #276: Hrp::parse_display accepts mixed-case input across write_str boundaries

## Finding

- title: Hrp::parse_display accepts mixed-case input across write_str boundaries
- severity: `medium`
- location: `src/primitives/hrp.rs:136-182`
- cwe: CWE-1023
- scanner: `llm-code-review`
- fingerprint: `81252bbb5f39df10f6bb4fc40d20127d29cf0a7979f21e30c9059b29cc00d87b`

## Description

`Hrp::parse_display` validates case inside `ByteFormatter::write_str`, but `has_lower`/`has_upper` are declared *inside* the method and reset on every call. When the supplied `Display` is delivered to the formatter as multiple segments (e.g. `format_args!("{}{}", "abc", "DEF")`, or any custom `Display` impl that issues two `write_str` calls), the per-call check is satisfied locally and mixed case across segments is never detected. `Hrp::parse` (which iterates a single `&str` in one pass) correctly rejects the same input, so the two parsers disagree.

The Hrp struct's own documentation (line 60) states that `buf` is "guaranteed not to be mixed-case." Calling `Hrp::parse_display(format_args!("{}{}", "abc", "DEF"))` returns `Ok` with buf containing the literal mixed-case bytes `abcDEF`, breaking that invariant. BIP-173 forbids mixed-case HRPs because the checksum is computed over the lowercase form, so values produced this way are downstream-uncashable in any honest decoder while still comparing equal (case-insensitively) to a properly-lowercase HRP in this crate's `PartialEq`. Any code path that allow-lists or denies HRPs after construction but then re-uses the original case (e.g., re-encoding, logging the address back to the user, feeding it to a different bech32 implementation) inherits the spec violation.

Impact: invariant break / spec-conformance bypass with downstream confused-deputy potential. Fix: hoist `has_lower`/`has_upper` to fields of `ByteFormatter` so they persist across `write_str` calls, mirroring `Hrp::parse`.

## Proof of Concept

```diff
--- a/src/primitives/hrp.rs
+++ b/src/primitives/hrp.rs
@@ -770,6 +770,18 @@ mod tests {
         let hrp = Hrp::parse_unchecked("!\x7f~");
         assert_eq!(hrp.as_bytes()[0], b'!');
         assert_eq!(hrp.as_bytes()[1], b'X');
         assert_eq!(hrp.as_bytes()[2], b'~');
     }
+
+    #[test]
+    fn parse_display_detects_mixed_case_across_writes() {
+        // The Hrp struct claims its buf is "guaranteed not to be mixed-case".
+        // When parse_display receives a Display that produces multiple
+        // write_str calls (e.g. format_args! with multiple args), the
+        // per-call has_lower/has_upper bookkeeping is reset between calls,
+        // so mixed case across segments is not detected and a mixed-case
+        // HRP is accepted in violation of BIP-173.
+        let result = Hrp::parse_display(format_args!("{}{}", "abc", "DEF"));
+        assert_eq!(result.unwrap_err(), Error::MixedCase);
+    }
 }

```

alexgrad42:
Move `has_lower` and `has_upper` flags from local variables inside `write_str` to fields on the `ByteFormatter` struct. This ensures that case validation persists across multiple `write_str` calls (e.g., when processing split `Display` arguments) instead of resetting on each boundary.

Fixes: #276

### DIFF
--- a/src/primitives/hrp.rs
+++ b/src/primitives/hrp.rs
@@ -131,12 +131,12 @@ impl Hrp {
             arr: [u8; MAX_HRP_LEN],
             index: usize,
             error: Option<Error>,
+            has_lower: bool,
+            has_upper: bool,
         }
 
         impl core::fmt::Write for ByteFormatter {
             fn write_str(&mut self, s: &str) -> fmt::Result {
-                let mut has_lower: bool = false;
-                let mut has_upper: bool = false;
                 for ch in s.chars() {
                     let b = ch as u8; // cast ok, `b` unused until `ch` is checked to be ASCII
 
@@ -151,17 +151,17 @@ impl Hrp {
                     }
 
                     if ch.is_ascii_lowercase() {
-                        if has_upper {
+                        if self.has_upper {
                             self.error = Some(MixedCase);
                             break;
                         }
-                        has_lower = true;
+                        self.has_lower = true;
                     } else if ch.is_ascii_uppercase() {
-                        if has_lower {
+                        if self.has_lower {
                             self.error = Some(MixedCase);
                             break;
                         }
-                        has_upper = true;
+                        self.has_upper = true;
                     };
                 }
 
@@ -181,7 +181,13 @@ impl Hrp {
             }
         }
 
-        let mut byte_formatter = ByteFormatter { arr: [0; MAX_HRP_LEN], index: 0, error: None };
+        let mut byte_formatter = ByteFormatter {
+            arr: [0; MAX_HRP_LEN],
+            index: 0,
+            error: None,
+            has_lower: false,
+            has_upper: false,
+        };
 
         write!(byte_formatter, "{}", data).expect("custom Formatter cannot fail");
         if byte_formatter.index == 0 {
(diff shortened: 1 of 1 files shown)


tcharding:
Needs unit test. Done as a separate patch last so review can verify the bug this fixes.

alexgrad42:
I've extracted the test into separate commit, please check

tcharding:
Thanks, still a couple things to fix please:

- The commit brief log (the title) should not be the same for both commits 
- The commit that adds the tests has to be second so that test pass on all commits. The during review someone will re-order the commits and check that the test fails without the fix.

Thanks

tcharding:
I'm reasonably sure this was mentioned to you on another PR. I'm also still uncertain how much of you work is human and how much LLM. If you are using an LLM you will need to configure it to remember stuff we mention in review otherwise it waists review time. Of course simply forgetting things if you are actually human is ok.

tcharding:
On further thought the tests are probably too verbose. Why not just use the one in the issue (witthout all the code comments).

alexgrad42:
I have rearranged the commit and cleaned the test. Note that example from Loupe in fact doesn't work, I've added a case to show this.

apoelstra (approved):
ACK a5312e727aa9e70c8c47989c93e3100c709b44fc; successfully ran local tests

PR #278, opened 2026-06-16
fix: make Hrp Hash case-insensitive to match PartialEq

### LINKED ISSUE #275: Hrp Hash impl is case-sensitive while PartialEq is case-insensitive

## Finding

- title: Hrp Hash impl is case-sensitive while PartialEq is case-insensitive
- severity: `medium`
- location: `src/primitives/hrp.rs:323-335`
- cwe: CWE-697
- scanner: `llm-code-review`
- fingerprint: `74a106c2abfc71d3decdd8142b2b3129a01d0d489400c7fac70733633edc63d3`

## Description

`Hrp::eq` (line 325) compares via `lowercase_byte_iter().eq(...)`, so two HRPs that differ only in case (e.g. `bc` and `BC`) compare equal. The `Hash` impl (line 334), however, hashes `self.buf` directly, which preserves the originally-parsed case. This violates the core `Hash`/`Eq` invariant from `core::hash::Hash`: "If two keys are equal, their hashes must also be equal."

Concretely: `Hrp::parse_unchecked("BC") == Hrp::parse_unchecked("bc")` is true, but their `Hash` outputs differ. Any `HashMap<Hrp, _>` / `HashSet<Hrp>` therefore exhibits silent logical failure — inserting one case and looking up the equivalent other case can miss, and inserting both can produce two entries that `Hrp::eq` claims are the same key. Because `Hrp` is public API, downstream code that uses it as a hash key (e.g. allow-lists keyed by HRP, dedup sets, caches keyed on the network identifier) is exposed to lookup misses that, depending on call-site semantics, become authorization bypasses or correctness vulnerabilities (the attacker-controlled half of the comparison can be uppercase, the trusted entry lowercase, or vice versa).

Fix: make `Hash` consume `self.lowercase_byte_iter()` (or hash a normalized prefix of `buf[..size]`) so equal HRPs always hash identically. The existing test `ordering_and_hash` only asserts non-zero hash and does not catch this contract violation.

## Proof of Concept

```diff
--- a/src/primitives/hrp.rs
+++ b/src/primitives/hrp.rs
@@ -770,6 +770,35 @@ mod tests {
         let hrp = Hrp::parse_unchecked("!\x7f~");
         assert_eq!(hrp.as_bytes()[0], b'!');
         assert_eq!(hrp.as_bytes()[1], b'X');
         assert_eq!(hrp.as_bytes()[2], b'~');
     }
+
+    #[test]
+    fn hash_consistent_with_eq() {
+        // The Hash trait requires: if a == b, then hash(a) == hash(b).
+        // Hrp::eq is case-insensitive (via lowercase_byte_iter) but Hrp's
+        // Hash impl hashes the raw buf, so equal HRPs that differ in case
+        // hash to different values. This breaks HashMap/HashSet correctness.
+        use core::hash::{Hash, Hasher};
+
+        struct Simple(u64);
+        impl Hasher for Simple {
+            fn finish(&self) -> u64 { self.0 }
+            fn write(&mut self, bytes: &[u8]) {
+                for &b in bytes {
+                    self.0 = self.0.wrapping_mul(31).wrapping_add(b as u64);
+                }
+            }
+        }
+
+        let a = Hrp::parse_unchecked("ABC");
+        let b = Hrp::parse_unchecked("abc");
+        assert_eq!(a, b);
+
+        let mut h1 = Simple(0);
+        let mut h2 = Simple(0);
+        a.hash(&mut h1);
+        b.hash(&mut h2);
+        assert_eq!(h1.finish(), h2.finish());
+    }
 }

```

alexgrad42:
Fixes the Hash implementation for the Hrp struct to iterate over lowercase bytes instead of hashing the raw internal buffer. This restores the core invariant where `a == b` implies `hash(a) == hash(b)`.

Fixes: #275

### DIFF
--- a/src/primitives/hrp.rs
+++ b/src/primitives/hrp.rs
@@ -779,4 +779,54 @@ mod tests {
         assert_eq!(hrp.as_bytes()[1], b'X');
         assert_eq!(hrp.as_bytes()[2], b'~');
     }
+
+    struct Simple(u64);
+
+    impl Hasher for Simple {
+        fn finish(&self) -> u64 { self.0 }
+        fn write(&mut self, bytes: &[u8]) {
+            for &b in bytes {
+                self.0 = self.0.wrapping_mul(31).wrapping_add(b as u64);
+            }
+        }
+    }
+
+    #[test]
+    fn hash_consistent_with_eq() {
+        let get_hash = |hrp: &Hrp| {
+            let mut hasher = Simple(0);
+            hrp.hash(&mut hasher);
+            hasher.finish()
+        };
+
+        let a = Hrp::parse_unchecked("ABC");
+        let b = Hrp::parse_unchecked("abc");
+        assert_eq!(a, b);
+        assert_eq!(get_hash(&a), get_hash(&b));
+
+        let mixed_1 = Hrp::parse_unchecked("aBcDeFg");
+        let mixed_2 = Hrp::parse_unchecked("AbCdEfG");
+        assert_eq!(mixed_1, mixed_2);
+        assert_eq!(get_hash(&mixed_1), get_hash(&mixed_2));
+
+        let num_1 = &Hrp::parse_unchecked("bc1");
+        let num_2 = &Hrp::parse_unchecked("BC1");
+        assert_eq!(num_1, num_2);
+        assert_eq!(get_hash(num_1), get_hash(num_2));
+
+        let short = Hrp::parse_unchecked("abc");
+        let long = Hrp::parse_unchecked("abcd");
+        assert_ne!(short, long);
+        assert_ne!(get_hash(&short), get_hash(&long));
+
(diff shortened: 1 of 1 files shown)


apoelstra:
I think this is wrong because if you hash two `Hrp`s this way you'll get the same hash as a concatenation of the HRPs.  See https://doc.rust-lang.org/std/hash/trait.Hash.html#prefix-collisions

The std impl for `[T]` adds a length prefix: https://doc.rust-lang.org/src/core/hash/mod.rs.html#934

alexgrad42:
I've added the length + extracted the test into separate commit, please check

apoelstra:
In ad99e2f85880ab55ce32872ad8247187f4833eb6:

now this has *two* length prefixes and does an unnecessary copy

PR #282, opened 2026-07-10
Fix pedantic clippy lints

Abeeujah:
This PR refactors a couple of data flow control and constructs regarding the `Option<T>` type, to be as idiomatic and ergonomic as the MSRV permits.

Commits are by Modules, but can be squashed if needed.


apoelstra (approved):
ACK 5b262ee77a20739a984b484e8af6b34e9f3af084; successfully ran local tests

PR #284, opened 2026-07-11
Stream unchecked SegWit encoder output

### LINKED ISSUE #273: Unchecked segwit encoders silently truncate output to MAX_STRING_LENGTH

## Finding

- title: Unchecked segwit encoders silently truncate output to MAX_STRING_LENGTH
- severity: `low`
- location: `src/segwit.rs:156-326`
- cwe: CWE-130
- scanner: `llm-code-review`
- fingerprint: `13d3aa43d4b893e86e39c983b74a34906382021fce5d7fc5561dabf1b8b8ed19`

## Description

The public `encode_*_unchecked` functions in `src/segwit.rs` (`encode_to_fmt_unchecked` line 143, `encode_lower_to_fmt_unchecked` line 156, `encode_upper_to_fmt_unchecked` line 196, `encode_to_writer_unchecked` line 237, `encode_lower_to_writer_unchecked` line 254, `encode_upper_to_writer_unchecked` line 296) all use the same pattern: a fixed `[0u8; MAX_STRING_LENGTH]` stack buffer (90 bytes) combined with `buf.iter_mut().zip(bytes).for_each(...)`. `zip` stops at the shorter iterator, so when the encoded length exceeds 90 characters (e.g. an HRP longer than 18 bytes combined with a max-length witness program), the iterator silently drops every byte beyond position 90 — including most of the 6-byte checksum — and the function still returns `Ok(())`.

In contrast, the streaming encoders in `src/lib.rs` (e.g. `encode_lower_to_writer`) flush the buffer in chunks, so they faithfully emit any length the caller passes. And `encode()` (line 103) pre-validates total length via `encoded_length()` before calling the unchecked path. The bug therefore only triggers when callers reach for the `*_unchecked` API directly (this is the no-alloc / writer path documented for QR-code use cases): a caller that validates the witness version and program length per the docstring's instructions, but is unaware that HRP length also affects encoded length, will receive a silently truncated "valid-looking" bech32 string with a corrupted checksum.

Impact is limited because the truncated string almost never satisfies the checksum, so downstream decoders reject it — but the surprise-free contract of a public API is that a successful `Ok` return wrote what was asked. A wallet that uses these writers to emit addresses into a UI or QR code could display a maimed string without any error, masking input-validation gaps in the caller.

Fix: stream into the writer in chunks (mirroring `encode_lower_to_writer` / `encode_upper_to_writer` in `src/lib.rs`), or count the bytes-iterator output and detect overflow rather than relying on `zip`.

## Proof of Concept

```diff
diff --git a/src/segwit.rs b/src/segwit.rs
--- a/src/segwit.rs
+++ b/src/segwit.rs
@@ -626,3 +626,18 @@
         assert_eq!(decode(address).unwrap_err(), DecodeError(SegwitHrpstringError::TooLong(91)));
     }
+
+    #[test]
+    fn encode_lower_to_fmt_unchecked_must_not_silently_truncate() {
+        // 22-char HRP + 1 separator + 1 witness version + 64 data + 6 checksum = 94 chars,
+        // which exceeds MAX_STRING_LENGTH (90). The encoder silently truncates because it
+        // writes into a fixed [u8; MAX_STRING_LENGTH] stack buffer using
+        // `buf.iter_mut().zip(bytes)`, which stops at the shorter iterator and drops the
+        // remaining bytes (including most of the checksum) with no error returned.
+        let hrp = Hrp::parse_unchecked("anhrpthatistwentychars");
+        let program = [0_u8; 40];
+        let mut got = String::new();
+        encode_lower_to_fmt_unchecked(&mut got, hrp, VERSION_1, &program)
+            .expect("encode_lower_to_fmt_unchecked");
+        assert_eq!(got.len(), 94, "encoded address was silently truncated to MAX_STRING_LENGTH");
+    }
 }

```

Abeeujah:
The unchecked encoders silently truncate output longer than the
fixed 90-byte stack buffer while still returning success.

Flush full buffers to the formatter or writer before continuing.
This preserves no-allocation encoding while ensuring unchecked
inputs are serialized completely.

closes #273

### DIFF
--- a/src/segwit.rs
+++ b/src/segwit.rs
@@ -159,29 +159,17 @@ pub fn encode_lower_to_fmt_unchecked<W: fmt::Write>(
     witness_version: Fe32,
     witness_program: &[u8],
 ) -> fmt::Result {
-    let mut buf = [0u8; MAX_STRING_LENGTH];
-    let mut pos = 0;
-
     let iter = witness_program.iter().copied().bytes_to_fes();
     match witness_version {
         VERSION_0 => {
             let bytes = iter.with_checksum::<Bech32>(&hrp).with_witness_version(VERSION_0).bytes();
-            buf.iter_mut().zip(bytes).for_each(|(dst, src)| {
-                *dst = src;
-                pos += 1;
-            });
+            write_bytes_to_fmt(fmt, bytes, false)?;
         }
         version => {
             let bytes = iter.with_checksum::<Bech32m>(&hrp).with_witness_version(version).bytes();
-            buf.iter_mut().zip(bytes).for_each(|(dst, src)| {
-                *dst = src;
-                pos += 1;
-            });
+            write_bytes_to_fmt(fmt, bytes, false)?;
         }
     }
-
-    let s = core::str::from_utf8(&buf[..pos]).expect("we only write ASCII");
-    fmt.write_str(s)?;
 
     Ok(())
 }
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK aa7582b3cee54d99c5818e6e937126934e9e3c99; successfully ran local tests

PR #302, opened 2026-09-11
  correction: reject invalid error magnitudes before iteration

### LINKED ISSUE #290: ErrorIterator::next reaches unreachable!() and aborts on some inputs
`ErrorIterator::next` hits an `unreachable!()` and aborts the process while computing BCH error correction. A short attacker-chosen string crashes any wallet or address-entry UI that offers "did you mean" correction on user-supplied text.

Version: bech32 0.12.0, default features (`alloc`, `std`). The minimum set is `default-features = false, features = ["alloc"]`.

## Reproducer

```rust
use bech32::primitives::decode::CheckedHrpstring;
use bech32::{Bech32m, CorrectableError};

fn main() {
    let s = "cbbc1yuglpryqdm";
    let err = CheckedHrpstring::new::<Bech32m>(s).unwrap_err();
    let ctx = err.correction_context::<Bech32m>().unwrap();
    let mut it = ctx.bch_errors().unwrap(); // Some, so within the singleton bound
    let _ = it.next();                      // panics
}
```

Same panic through the top-level API:

```rust
use bech32::{Bech32m, CorrectableError};

fn main() {
    bech32::segwit::decode("cbbc1yuglpryqdm")
        .unwrap_err()
        .correction_context::<Bech32m>()
        .unwrap()
        .bch_errors()
        .unwrap()
        .next();
}
```

## Observed vs expected

Observed: the process aborts at `unreachable!("error guaranteed to  lie in base field")`.

Expected: the rustdoc for `bch_errors()` says it returns `None` when there are too many errors to correct and otherwise yields `(usize, Fe32)` pairs. `Corrector`, `ErrorIterator`, and `Iterator::next` document no panic. The rustdoc on `ErrorIterator` anticipates adversarial input and tells the caller to reject such strings, which the caller cannot do if the library aborts first. Either `bch_errors()` should return `None` here, or `next()` should return `None` or an out-of-range location.

## Root cause

`src/primitives/correction.rs:302`. `next` computes the Forney error value in the extension field `Fe1024`, then narrows it with `ret.try_into()`. When Berlekamp-Massey converges on a spurious error locator, the value carries a non-zero degree-1 coefficient, `TryFrom<Fe32Ext<2>> for Fe32` returns `Err`, and the `Err(_)` arm panics.

## Scope

The panic needs two conditions: `bch_errors()` returns `Some`, so the syndrome sits within the singleton bound, and Berlekamp-Massey converges on a locator matching no real correctable error pattern. `aa1aagz20up2hk8e` also panics on the `Bech32m` path above. About 0.05% of random strings trigger the panic, near 40 in every 80,000.

ANAMASGARD:
Fixes #290.

-  Malformed input can produce BCH correction magnitudes outside Fe32, causing `ErrorIterator::next()` to panic. 
 
 - Validated all candidates in `bch_errors()` before returning an iterator, rejecting invalid magnitudes and zero denominators through a shared fallible Forney  calculation.

  - Restored the iterator after validation so valid corrections, including zero-valued erasures, remain available. 
  - Added regression tests for  both reported inputs and extend fuzz coverage to fully consume correction iterators.


apoelstra:
Superceded by #304.

PR #307, opened 2026-09-16
Various iterator length computation fixes

apoelstra:
Our `size_hint` and `len` methods for our `BytesToFes` and `FesToBytes` iterators were easy to make overflow, and also they were wrong.

Unlike for slices, where you can't really get crazy overflowing values without massive allocations, it's easy to make iterators with huge lengths. `iter::repeat(whatever).take(N)` will do it. On modern rustcs (sadly not our MSRV) this construction is even an `ExactSizeIterator` somehow. So it's plausble that normal callers of these APIs would bump into these issues.

Fix all these issues, and add exhaustive tests for correctness for small values

Removes the `ExactSizeIterator` impl for `BytesToFes`. As much as it pains me, we are not able to implement this trait because an arbitrary byte iterator will yield more fes than bytes, so if the original iterator had close to `usize::MAX` many items, then the fe-converted one will have an overflowing length that we cannot report. To compensate, I added an `exact_size` inherent method that will give you the old value.

Fixes https://github.com/project-loupe/audit-rust-bech32/issues/14, https://github.com/project-loupe/audit-rust-bech32/issues/20, https://github.com/project-loupe/audit-rust-bech32/issues/22, https://github.com/project-loupe/audit-rust-bech32/issues/29, https://github.com/project-loupe/audit-rust-bech32/issues/32


tcharding (approved):
ACK 0b643ab0b7d883ac950484c3fef7056ec570c1b2

tcharding:
Seems sane. I only really looked at the test code.

src/lib.rs
```diff
@@ -749,6 +759,26 @@ mod tests {
...
+        // Define a checksum which is the same as bech32 except we claim that it has usize::MAX
+        // length.
+        enum OverflowChecksum {}
+        impl Checksum for OverflowChecksum {
+            type MidstateRepr = <Bech32 as Checksum>::MidstateRepr;
+            type CorrectionField = <Bech32 as Checksum>::CorrectionField;
+            const ROOT_GENERATOR: Self::CorrectionField = <Bech32 as Checksum>::ROOT_GENERATOR;
+            const ROOT_EXPONENTS: core::ops::RangeInclusive<usize> = 24..=26;
+            const CODE_LENGTH: usize = usize::MAX; // lmao
+            const CHECKSUM_LENGTH: usize = usize::MAX; // lmao
```
tcharding, line 774:
```suggestion
            const CODE_LENGTH: usize = 1023;
            const CHECKSUM_LENGTH: usize = usize::MAX; // lmao
```

So that its more clear that the test name ties to what is overflowing.

apoelstra: I considered this, but it's a little incoherent to have a checksum length that exceeds the code length. (Though this whole "checksum" is incoherent so does it really matter?)

apoelstra:
On 0b643ab0b7d883ac950484c3fef7056ec570c1b2 successfully ran local tests

PR #310, opened 2026-09-18
A couple more overflow fixes

apoelstra:
Apparently I'm not allowed to get to zero, because PL opened three more issues in the night before I was able to merge my "fix all the PL errors" PRs.

Fix two of them, which are more-or-less straightforward "unchecked arithmetic" errors. The other one probably I will promote to an issue here because it's a bit math-heavy and I'm not sure yet how I want to handle it.

Fixes https://github.com/project-loupe/audit-rust-bech32/issues/39, https://github.com/project-loupe/audit-rust-bech32/issues/40


apoelstra:
On 664bc4ef653adc7ef142562625c2b49143061335 successfully ran local tests

